PR #1, opened 2019-03-27
Operator functionality, deployment and tests

yanirq:
Node maintenance operator functionality and tests as described in  [README](https://github.com/kubevirt/node-maintenance-operator/blob/master/README.md).

In addition,  when a node is put into maintenance,  the operator will also add a special taint `kubevirt.io/drain:NoSchedule` to drained nodes that will indicate if VMs should be live migrated.
VMs will act accordingly using tolerations for that taint, see https://github.com/kubevirt/kubevirt/pull/2091


deploy/crds/kubevirt_v1alpha3_nodemaintenance_crd.yaml
```diff
@@ -0,0 +1,33 @@
...
+        apiVersion:
+          type: string
+        kind:
+          type: string
+        metadata:
+          type: object
+        spec:
+          type: object
+        status:
+          type: object
```
MarSik: Should we be a bit more specific here? Or is this enough?

yanirq: Actually this was generated by the operator-sdk,  I think for now its enough but ill give it a second look.

pkg/controller/nodemaintenance/drain.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	_, err := waitForDelete(pods, 1*time.Second, globalTimeout, false, getPodFn)
+	return err
+}
+
+func evictPods(r *ReconcileNodeMaintenance, pods []corev1.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*corev1.Pod, error)) error {
+	returnCh := make(chan error, 1)
+
+	for _, pod := range pods {
+		go func(pod corev1.Pod, returnCh chan error) {
```
rmohr: I know this is coming from kubectl directly, but this looks like it has the potential to create a lot of traffic is more than one node with many pods are put into maintenance. If you run evict on three or four nodes in parallel and there is a non-standard situation, you have a good chance to do 1000 requests per second and either create a lot of load or get rate-limited.

Here we have a go-routine which does the delete for every pod in parallel + periodic polls on the pod state per pod in parallel. 

Does it have more than one worker on the controller?

yanirq: we have one worker

build/bin/user_setup
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -x
+
+# ensure $HOME exists and is accessible by group 0 (we don't know what the runtime UID will be)
+mkdir -p ${HOME}
+chown ${USER_UID}:0 ${HOME}
+chmod ug+rwx ${HOME}
+
+# runtime user will need to be able to self-insert in /etc/passwd
+chmod g+rw /etc/passwd
```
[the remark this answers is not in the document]
yanirq: this is auto generated from operator-sdk

build/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ENV OPERATOR=/usr/local/bin/node-maintenance-operator \
+    USER_UID=1001 \
+    USER_NAME=node-maintenance-operator
+
+# install operator binary
+COPY build/_output/bin/node-maintenance-operator ${OPERATOR}
+
+COPY build/bin /usr/local/bin
+RUN  /usr/local/bin/user_setup
```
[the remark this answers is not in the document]
yanirq: this is auto generated from operator-sdk

build/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ENTRYPOINT ["/usr/local/bin/entrypoint"]
```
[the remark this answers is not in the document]
yanirq: this is auto generated from operator-sdk

yanirq:
> Did a first iteration. I am a little bit concerned that the implementation endangers the cluster stability or gets rate-limited pretty fast.
> 
> If my observations are correct, I would rather recommend an informer based on informers with e.g. 3 workers. Each of these workers can do X deletes at max in parallel. After they were invoked, it waits until watches on the resources report all deletes in the cache, then go on with the next round of deletions. That is how other controller like this work.

@rmohr  Added a pod informer as discussed to get and check if pods were deleted from the evicted node.

pkg/controller/nodemaintenance/drain.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil
+	}
+
+	policyGroupVersion, err := drain.CheckEvictionSupport(r.drainer.Client)
+	if err != nil {
+		return err
+	}
+
+	getPodFn := func(name string) (bool, error) {
+		_, exists, err := r.podInformer.GetStore().GetByKey(name)
```
rmohr: Here you have to pass `namespace + "/" + name`, otherwise you will not be able to find the pod.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	return node, nil
+}
+
+func (r *ReconcileNodeMaintenance) startPodInformer(node *corev1.Node, stop <-chan struct{}) error {
+	fieldSelector := fields.SelectorFromSet(fields.Set{"spec.nodeName": node.Name})
+
+	lw := cache.NewListWatchFromClient(
+		r.drainer.Client.CoreV1().RESTClient(),
+		"pods",
+		node.Namespace,
```
rmohr: nodes have no namespaces, you can use `NamespaceAll`, that's a constant which by accident has the same value like `node.namespace`, which is `""`.

pkg/controller/nodemaintenance/drain.go
```diff
@@ -0,0 +1,190 @@
...
+	return nil
+}
+
+func drainPods(r *ReconcileNodeMaintenance, nodeName string) error {
+	list, errs := r.drainer.GetPodsForDeletion(nodeName)
+
+	if errs != nil {
+		return utilerrors.NewAggregate(errs)
+	}
+
```
rmohr: You will have to start the informer exactly here. There is a highly theoritical chance otherwise that  the cache does not see a new pod which got added between the list call here and the informer start.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil, err
+	} else if err != nil {
+		log.Error(err, fmt.Sprintf("Failed to get Node %s: %v\n", nodeName, err))
+		return nil, err
+	}
+	return node, nil
+}
+
+func (r *ReconcileNodeMaintenance) startPodInformer(node *corev1.Node, stop <-chan struct{}) error {
+	fieldSelector := fields.SelectorFromSet(fields.Set{"spec.nodeName": node.Name})
```
rmohr: You will also have to select based on phases, you only want pods which are still running.

rmohr: ah nonsense.

PR #2, opened 2019-04-08
Add node name under Spec

### LINKED ISSUE #3: Use a finalizer to untaint the node when the CR is deleted
Right now the taints would stay in place after the node is in maintenance. Is it expected to remove the taints independent of the controller, or should they be removed when the CR is deleted?

yanirq:
Changed scheme for NodeMaintenance CR to support the following structure:

apiVersion: kubevirt.io/v1alpha1
kind: NodeMaintenance
metadata:
  name: nodemaintenance-xyz
spec:
  nodeName: node02
  reason: "Test node maintenance"

fixes #3

### DIFF
--- a/pkg/apis/kubevirt/v1alpha1/nodemaintenance_types.go
+++ b/pkg/apis/kubevirt/v1alpha1/nodemaintenance_types.go
@@ -4,9 +4,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	NodeMaintenanceFinalizer string = "foregroundDeleteNodeMaintenance"
+)
+
 // NodeMaintenanceSpec defines the desired state of NodeMaintenance
 // +k8s:openapi-gen=true
 type NodeMaintenanceSpec struct {
+	// Node name to apply maintanance on/off
+	NodeName string `json:"nodeName"`
+	// Reason for maintanance
 	Reason string `json:"reason,omitempty"`
 }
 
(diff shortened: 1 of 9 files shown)


pkg/apis/kubevirt/v1alpha1/nodemaintenance_types.go
```diff
@@ -4,9 +4,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	NodeMaintenanceFinalizer string = "foregroundDeleteNodeMaintenance"
```
rmohr: Ah great here a finalizer for #3 is introduced. :+1:

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -148,15 +148,36 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+			if err := r.client.Update(context.TODO(), instance); err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+	} else {
+		// The object is being deleted
+		if containsString(instance.ObjectMeta.Finalizers, kubevirtv1alpha1.NodeMaintenanceFinalizer) {
+			// remove our finalizer from the list and update it.
+			instance.ObjectMeta.Finalizers = removeString(instance.ObjectMeta.Finalizers, kubevirtv1alpha1.NodeMaintenanceFinalizer)
+			if err := r.client.Update(context.Background(), instance); err != nil {
```
rmohr: You will have to update the operator status after you uncordoned the node. Otherwise you would not retry to uncordon the node if you have an error when calling the api after this call. Normally it is like this:

1. check if I have something to do
2. do it
3. update the status of the main CR

This way you will retry in case you have any kind of errors until you reached your target. If you do it this way:

1. check if I have something to do
2. update the status of the main CR
3. do it

you would never retry 3. if it fails the first time because you already updated the status.

PR #17, opened 2019-05-16
Fix kubevirt taint addition/removal

yanirq:
This comes to fix the Issue: Running node maintenance adds NoSchedule taints and evicted all pods, as expected. Sometimes, after deleting this maintenance job, NoSchedule taint was not removed, so new pods can't be created on this node.

It seems that the problem appears after the drain process reaches its timeout (1 minute) and the controller reconciles again. In the following reconcile loops the taint is being added again although already added.

This PR comes to fix the unneeded additions of the taint to a node.


yanirq:
@rmohr  Currently there is still a problem with VM migration. with this fix, the controller will not add the taint again (will not queue an additional update) , but although i can see in the node description that the taint is added, the vm pod is not evicted due to PDB that still exist.  
Still trying to reach a conclusion here.

/cc @fabiand @MarSik

yanirq:
@rmohr Currently there is still a problem with VM migration. with this fix, the controller will not add the taint again (will not queue an additional update) , but although i can see in the node description that the taint is added, the vm pod is not evicted due to PDB that still exist.
Still trying to reach a conclusion here.

/cc @fabiand @MarSik @rmohr

yanirq:
> Conceptually this looks fine to me, but I know almost nothing about the interaction and patching intricacies so @rmohr should approve as well before merging.

There is still the problem of un-evicted VM pod currently with this fix.

rmohr:
> There is still the problem of un-evicted VM pod currently with this fix.

@yanirq on which platform are you testing? There are two bugs in k8s/openshift which can cause issues:

 * on openshift with crio pods get stuck in terminating because crio does not send a signal to the containers
 * Until k8s 1.13 there was a bug in PDBs where PDBs were protecting terminated pods, which it obviously shouldn't.

rmohr:
> * on openshift with crio pods get stuck in terminating because crio does not send a signal to the containers

and we have a workaround in kubevirt itself starting from 0.17.

pkg/controller/nodemaintenance/taint.go
```diff
@@ -40,29 +40,31 @@ func AddOrRemoveTaint(clientset kubernetes.Interface, node *corev1.Node, add boo
...
-	if patchErr != nil {
-		log.Error(err, fmt.Sprintf("%s taint %s was not applied on Node: %s", taint.Key, taintStr, node.Name))
+		newData, err := json.Marshal(freshNode)
+		if err != nil {
+			return err
+		}
+
+		patchBytes, patchErr := strategicpatch.CreateTwoWayMergePatch(oldData, newData, node)
+		if patchErr == nil {
+			_, err = client.Patch(node.Name, types.StrategicMergePatchType, patchBytes)
```
rmohr: I just checked, tains don't have a "merge key" for the strategic merge patch. Tains have a "Key" field, but it does not act as key for the patch operation. Can you change the patch type to json patch?

This will require a patch operation which consists of a `test` and a `replace` operation. Here an example: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/watch/vmi.go#L370.

If the list can also be `nil`, you may have to check for  that too and do an `add` if it is nil.

yanirq:
Added taint checks on node before each operation that adds/removes taints so there will be no conflicts due to race conditions.
Patching taints to node is now done with JSONPatchType
(/cc @rmohr) 

CI tests should also include k8s version 1.13+  since TaintNodesByCondition feature gate is enabled by default only for versions higher than 1.12.  these will be included in a different PR.

yanirq:
@davidvossel  Waiting on node objects - now removed and the flow is more idempotent.
The issue with taints appeared to be created due to the node lifecycle controller which adds a `node.kubernetes.io/unschedulable:NoSchedule` taint after we cordon the node. 
Further more, the Patch that was applied to remove the kubevirt taint used op: replace which seemed to have issues with consistency. 
Now, Instead of using replace the code uses remove which seems to do the job.
Do see any issues with the current fix ?

p.s
Usage of node informer if needed and addition of unit tests will be done in a different PR.

yanirq:
@davidvossel  I don't think we can remove the `update` rbac just yet.
An update operation can still occur in: 
https://github.com/kubevirt/node-maintenance-operator/blob/a6facb49dbfde90fdf76ec8f8066542f9dc97538/pkg/controller/nodemaintenance/drain.go#L25 
--->
https://github.com/kubevirt/node-maintenance-operator/blob/a6facb49dbfde90fdf76ec8f8066542f9dc97538/vendor/k8s.io/kubernetes/pkg/kubectl/drain/cordon.go#L90-L96

PR #20, opened 2019-05-26
Add/Remove NodeUnschedulableTaint on maintanance creation/deletion.

yanirq:
This PR comes to deal with a race condition caused by node lifecycle controller which adds/removes a `node.kubernetes.io/unschedulable:NoSchedule` taint after we cordon/uncordon the node
and by that causes the `node.kubernetes.io/unschedulable:NoSchedule` not to be added/removed (sometimes - according to the race)  

For that reason, also the `node.kubernetes.io/unschedulable:NoSchedule` taint will be added/removed directly on/from the node to avoid collisions with the kubernetes node lifecycle controller

### DIFF
--- a/pkg/controller/nodemaintenance/nodemaintenance_controller.go
+++ b/pkg/controller/nodemaintenance/nodemaintenance_controller.go
@@ -187,14 +187,12 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	}
 
 	// Cordon node
-	if err := runCordonOrUncordon(r, node, true); err != nil {
+	err = AddOrRemoveTaint(r.drainer.Client, node, true)
+	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	// Add kubevirt live migration taint on the node
-	err = AddOrRemoveTaint(r.drainer.Client, node, true)
-
-	if err != nil {
+	if err = runCordonOrUncordon(r, node, true); err != nil {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


yanirq:
ci test please

yanirq:
ci test please

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -187,14 +187,12 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	}
 
 	// Cordon node
```
ksimon1: Maybe move this comment to line 194?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -214,16 +212,17 @@ func (r *ReconcileNodeMaintenance) stopNodeMaintenance(nodeName string) error {
 	if err != nil {
 		return err
 	}
+
 	// Uncordon the node
```
ksimon1: the same

pkg/controller/nodemaintenance/nodemaintenance_controller.go
yanirq: I moved it intentionally from 194 above to mark the entire process of uncordon - adding or removing the taints is basically uncordoning/cordoning the node

pkg/controller/nodemaintenance/nodemaintenance_controller.go
yanirq: same comment as above

PR #26, opened 2019-07-15
Add status fields and updates to maintenance CR

### LINKED ISSUE #16: Report maintenance status to CR status field
It would be nice if maintenance status is reported into CR status field. The CR represents a request that we want to put node under maintenance but we do not know when the node is actually ready (pods have been evicted etc).

yanirq:
**Status** added to maintenance CR's and will contain:
   #Phase is the represtation of a maintenanace progress (Running,Succeeded)
      **phase** : Running|Succeeded
  #LastError represents the latest error if any in the latest reconciliation
    **lastError**:
  #PendingPods are pods that failed to be evicted in the latest reconciliation
    **pendingPods**: list of pod objects

The fields will be updated according to the state by the controller in every reconciliation round.
Fixes #16


yanirq:
/cc @jtomasek @fabiand @cynepco3hahue @davidvossel  
I would like to have an initial review about the status structure .
Thanks

deploy/crds/nodemaintenance_crd.yaml
```diff
@@ -17,23 +16,51 @@ spec:
...
             reason:
+              description: Reason for maintanance
               type: string
           required:
           - nodeName
           type: object
         status:
           properties:
-            phase:
+            completed:
```
davidvossel: Was a Status.Phase enum considered to indicate things like Running|Succeeded|Failed?

That would be more consistent with how many of our other apis (and the k8s api) is structured.

deploy/crds/nodemaintenance_crd.yaml
```diff
@@ -17,23 +16,51 @@ spec:
...
+              description: Completed represents the state of maintenance. True if
+                completed , false otherwise
+              type: boolean
+            errorMessage:
```
davidvossel: Could we use Conditions to indicate specific error conditions along with human readable string?

deploy/crds/nodemaintenance_crd.yaml
```diff
@@ -17,23 +16,51 @@ spec:
...
+                during the latest reconciliation, false otherwise
+              type: boolean
+            pendingPods:
+              description: PendingPods are pods that failed to be evicted in the latest
+                reconciliation
+              items:
+                type: object
+              type: array
+            startTime:
+              description: StartTime contains the Date and time at which the maintenance
```
davidvossel: Is this very much different from the CreationTimestamp we already get in the metadata section?

yanirq: I was contemplating about this part, what is actually the start time.
Is it when the CR created (which makes sense) or is it when we actually started reconciling on it.
I now tend to remove this field and rely on the creationTimestamp.
@jtomasek  any thoughts on that from your perspective as a UI consumer ?

deploy/crds/nodemaintenance_crd.yaml
yanirq: Well there was a debate whether we should still use conditions or not and i refrained from using condition till now.  how critical is it to use conditions instead of a error string field that would indicate the last error ?

deploy/crds/nodemaintenance_crd.yaml
yanirq: I can agree to that.  
I'll remove the completed boolean field in the next commit and return phase with these values.
one thing to keep in mind is that this field can move from Failed to Running/Succeeded in sequential reconciliation loops.

deploy/crds/nodemaintenance_crd.yaml
jtomasek: I'd consider the maintenance as initiated once the CR is created, so yes `creationTimestamp` should be good enough.

yanirq:
ci test please

deploy/crds/nodemaintenance_crd.yaml
davidvossel: probably not a big deal

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -265,3 +279,22 @@ func (r *ReconcileNodeMaintenance) StartPodInformer(node *corev1.Node, stop <-ch
 
 	return nil
 }
+
+func (r *ReconcileNodeMaintenance) reconcileAndError(nm *kubevirtv1alpha1.NodeMaintenance, result reconcile.Result, err error) (reconcile.Result, error) {
+	//should always be false by default unless completed
+	nm.Status.Phase = kubevirtv1alpha1.MaintenanceFailed
```
davidvossel: Phase=Success|Failed need to be finalized states.

Is it possible for the phase to be set to Failed here yet at some point get retied and later on get set to Success? If so, that needs to change.

Conditions exist to represent these transient states where a CR might be in the Running Phase yet be temporarily failing on a transient error that will result in a retry later on.

yanirq: @davidvossel  good point.

> Is it possible for the phase to be set to Failed here yet at some point get retied and later on get set to Success?

Yes, it is possible with the current implementation. Since there is no timeout/time buffer field - we do not have a finalized state for failed. the controller will just keep trying to reconcile and process the maintenance CR until its deleted.

> Conditions exist to represent these transient states where a CR might be in the Running Phase yet be temporarily failing on a transient error that will result in a retry later on.

Conditions do look like the better way to express the transient errors if there is no refrain of still using them (also in future versions)
If so, then i would remove the LastError field that states the last transient error and replace it with conditions.

yanirq: @davidvossel  Fail status now removed

PR #31, opened 2019-08-07
Fix CI dependecies and flow

yanirq:


yanirq:
@gbenhaim  @danielBelenky  can you guys say why the CI is failing on python3 ? i tired several options here

davidvossel:
> @gbenhaim @danielBelenky can you guys say why the CI is failing on python3 ? i tired several options here

does automation/test.sh need to install python3?

yanirq:
> > @gbenhaim @danielBelenky can you guys say why the CI is failing on python3 ? i tired several options here
> 
> does automation/test.sh need to install python3?

dnf/yum commands inside the script cannot be found when attempted to add:
yum install -y python3 for example

automation/check-patch.k8s-1.11.0.packages
```diff
@@ -1,2 +1,3 @@
 docker
 make
+python3
```
gbenhaim: should be `python36` (also in the other *package files)

PR #34, opened 2019-08-18
Fix unwanted taint removals

yanirq:
Fix removal of existing taints that are not related to the maintenance operation while putting node in and out of maintenance .

**Problem description** 
Starting maintenance on a node should add `node.kubernetes.io/unschedulable:NoSchedule` and `kubevirt.io/drain:NoSchedule` taints to it , but in doing so it removed all other taints that pre existed prior to maintenance.
The same happened when stopping maintenance on a node.


yanirq:
ci test please

yanirq:
ci test please

yanirq:
> LGTM, but please elaborate a bit more on the commit message -few lines are sufficient- what happened and what this patch fixes.
> We can learn this from code itself, but a short intro saves time.

Ack.  Added a problem description to the PR.

PR #38, opened 2019-09-04
Close eviction routine on reconciliation

yanirq:
**Problem Description** 
For each reconciliation of the maintenance controller a go routine was created to call pod evictions.
The problem was that when a new reconciliation round has started these go routines kept running - trying to evict the same pod/s simultaneously 

**Solution**
Every time a reconciliation round ends (currently hard coded for 1 minute) 
The eviction go routine will be closed and a new one is opened. Meaning only one eviction routine for each reconciliation.


yanirq:
ci please test

yanirq:
ci please test

PR #49, opened 2019-12-25
Move to use Prow

gbenhaim:
- Since we added Prow jobs, we don't need stdci anymore.
- Add OWNERS file, so the maintainers will be notified when there is a
  new PR to review + be able to approve changes.

**Release note:**

```
None
```

Signed-off-by: gbenhaim <galbh2@gmail.com>


yanirq:
@gbenhaim  @ksimon1  I think we should add okd 4.3 and remove k8 1.11 in another patch later on

gbenhaim:
@yanirq can you please try to add the `approved` label by commenting "/approve"
If it won't work (since this PR adds the OWNERS file), this PR should be merged manually,

PR #50, opened 2020-02-04
WIP: Update NMO based on provisional upstream enhancement proposal

beekhof:


pkg/controller/machinemaintenance/machinemaintenance_controller.go
```diff
@@ -0,0 +1,491 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+
+	drain "k8s.io/kubectl/pkg/drain"
```
yanirq: you will need to vendor this package in (e.g. run go dep ensure) 
This is under the presumption it exists.  I would suggest checking the other dependencies under Gopkg.toml (if they need re-alignment to a higher k8s version

yanirq:
@beekhof  I wouldn't hurry to remove `pkg/controller/nodemaintenance/taint.go`  this is not something drain package provides.
If we will want to add/remove taints due to a maintenance operator (taints we express in the CR for example) we can use this functionality in the controller. 
Maybe it ok to remove just the hardcoded kubevirt taint for now.

pkg/controller/add_nodemaintenance.go
```diff
@@ -1,10 +1,10 @@
 package controller
 
 import (
-	"kubevirt.io/node-maintenance-operator/pkg/controller/nodemaintenance"
+	"kubevirt.io/node-maintenance-operator/pkg/controller/machinemaintenance"
```
yanirq: why is it so important to rename this to machinemaintenance ? (it kinds of defies the repository's name and purpose since we are operating on k8s nodes - unless the intent is more then just renaming the package here)

beekhof:
> @beekhof I wouldn't hurry to remove `pkg/controller/nodemaintenance/taint.go` this is not something drain package provides.
> If we will want to add/remove taints due to a maintenance operation (taints we express in the CR for example) we can use this functionality in the controller.
> Maybe it ok to remove just the hardcoded kubevirt taint for now.

Do we need the ``NodeUnschedulableTaint`` when we have `drain.RunCordonOrUncordon()` though?

beekhof:
@MichaelMoser  if you look at the tests executed by kubevirt-bot, you'll notice that this code does not compile for the reason @yanirq mentioned.  Please fix

```
---> Running in e8adcaa47375
pkg/controller/nodemaintenance/nodemaintenance_controller.go:29:2: cannot find package "k8s.io/kubectl/pkg/drain" in any of:
	/go/src/kubevirt.io/node-maintenance-operator/vendor/k8s.io/kubectl/pkg/drain (vendor tree)
	/usr/local/go/src/k8s.io/kubectl/pkg/drain (from $GOROOT)
	/go/src/k8s.io/kubectl/pkg/drain (from $GOPATH)
The command '/bin/sh -c CGO_ENABLED=0 GOOS=linux go build -o /node-maintenance-operator kubevirt.io/node-maintenance-operator/cmd/manager' returned a non-zero code: 1
Makefile:68: recipe for target 'container-build-operator' failed
```

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
 		return err
 	}
+	tmpDurationInSeconds := requestedMaintenenceWindowSeconds
+	lease := &coordv1beta1.Lease{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nodeName,
+			Namespace: corev1.NamespaceNodeLease,
+		},
+		Spec: coordv1beta1.LeaseSpec{
+			HolderIdentity:       &nodeName,
```
beekhof: This looks wrong.

`nodeName` is the target of the lease, this controller is the holder.
you should be putting something unique to this controller (probably "node-maintenance") in this field.

I think I also mentioned adding to the Node an annotation pointing to this lease object?
This is a pattern used elsewhere for establishing relationships between Nodes/Machines/Hosts.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
-	//The VirtualMachineInstance Pod does use emptryDir volumes,
-	//however the data in those volumes are ephemeral which means it is safe to delete after termination.
-	r.drainer.DeleteLocalData = true
+			}
 
-	//Ignore DaemonSet-managed pods.
-	//This is required because every node running a VirtualMachineInstance will also be running our helper DaemonSet called virt-handler.
-	//This flag indicates that it is safe to proceed with the eviction and to just ignore DaemonSets.
-	r.drainer.IgnoreAllDaemonSets = true
+		case LeaseStatusActiveOwnedByMeAboutToExpire:
```
beekhof: This only makes sense if we initially set the lease time to a value less than the requested value (but greater than the loop interval and worst case execution time).  If we cannot think of a reason to do that, we should drop this code.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
+			}
+			fallthrough
 
-	return nil
-}
+		case LeaseStatusActiveOwnedByMe:
+			{
 
-var _ reconcile.Reconciler = &ReconcileNodeMaintenance{}
+				if statusAnnoData.status != NodeStateCordon && statusAnnoData.status != NodeStateEvicting {
```
beekhof: Ok its illegal, but can we do about it?
Blocking until the lease expires doesn't seem like a great option

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
-// ReconcileNodeMaintenance reconciles a NodeMaintenance object
-type ReconcileNodeMaintenance struct {
-	// This client, initialized using mgr.Client() above, is a split client
-	// that reads objects from the cache and writes to the apiserver
-	client      client.Client
-	scheme      *runtime.Scheme
-	drainer     *drain.Helper
-	podInformer cache.SharedInformer
-}
+		case LeaseStatusActiveOwnedByDifferentNode:
```
beekhof: I don't understand, why would nodeA have a lease on nodeB?

This is a per-node lease, not a per-cluster one.
Owned by another controller (ie. machine healthcheck or MCO) is valid though

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
-// Reconcile reads that state of the cluster for a NodeMaintenance object and makes changes based on the state read
-// and what is in the NodeMaintenance.Spec
-// Note:
-// The Controller will requeue the Request to be processed again if the returned error is non-nil or
-// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
-func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithFields(log.Fields{"Request.Namespace": request.Namespace, "Request.Name": request.Name})
-	reqLogger.Info("Reconciling NodeMaintenance")
+				err := fmt.Errorf("lease got acquired by someone while request was active. node: %s", request.Name)
+				return reconcile.Result{}, err
```
beekhof: Nothing we should do here?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
-		return reconcile.Result{}, err
-	}
 
-	// Add finalizer when object is created
-	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
-		if !ContainsString(instance.ObjectMeta.Finalizers, kubevirtv1alpha1.NodeMaintenanceFinalizer) {
-			instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, kubevirtv1alpha1.NodeMaintenanceFinalizer)
-			if err := r.client.Update(context.TODO(), instance); err != nil {
-				return r.reconcileAndError(instance, err)
+		if drainer, kclient, err := r.initDrainer(); err == nil {
```
beekhof: Why create this each time?

pkg/controller/nodemaintenance/taint.go
```diff
@@ -1,102 +0,0 @@
-package nodemaintenance
```
beekhof: As per @yanirq, we might want to hold off deleting this file.
Especially if kubevirt is relying on the existence of `kubevirt.io/drain`

pkg/controller/machinemaintenance/machinemaintenance_controller.go
MoserMichael: added to vendor.

yanirq:
> > @beekhof I wouldn't hurry to remove `pkg/controller/nodemaintenance/taint.go` this is not something drain package provides.
> > If we will want to add/remove taints due to a maintenance operation (taints we express in the CR for example) we can use this functionality in the controller.
> > Maybe it ok to remove just the hardcoded kubevirt taint for now.
> 
> Do we need the `NodeUnschedulableTaint` when we have `drain.RunCordonOrUncordon()` though?

@beekhof  We do, especially if kubevirt still uses that to signal migration.
Having said that, its worth checking again what is the mechanism now and if KubeVirt has started to use eviction hooks. 
Also adding the unschedulable taint was done for backward compatibility reasons (If i remember correctly in some versions of k8s you have this taint added automatically and in some you didn't - need to double check that)

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
+
+// Reconcile monitors Nodes and creates the MachineRemediation object when the node has reboot annotation.
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+
+	r.reqLogger = log.WithFields(log.Fields{"Request.Namespace": request.Namespace, "Request.Name": request.Name})
+	r.reqLogger.Info("Reconciling ReconcileNodeMaintenance")
+	r.reqLogger.Infof("Reconciling node %s\n", request.Name)
```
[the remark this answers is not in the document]
yanirq: I agree on changing that to Debug messages (something i was thinking to do in the past but didn't get to it)

MoserMichael:
> @beekhof I wouldn't hurry to remove `pkg/controller/nodemaintenance/taint.go` this is not something drain package provides.
> If we will want to add/remove taints due to a maintenance operation (taints we express in the CR for example) we can use this functionality in the controller.
> Maybe it ok to remove just the hardcoded kubevirt taint for now.

brought it back.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -565,7 +565,11 @@ func (r *ReconcileNodeMaintenance) cancelEviction(drainer *drain.Helper, nodeNam
 	if len(pods) != 0 {
 		// cancel the move
 		for _, pod := range pods {
-			return drainer.Client.PolicyV1beta1().Evictions(pod.Namespace).Evict(nil)
+			err := drainer.Client.PolicyV1beta1().Evictions(pod.Namespace).Evict(nil)
```
beekhof: Have you tested this or found documentation that `Evict()` behaves like this?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -165,163 +168,215 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	nodeName := node.ObjectMeta.Name
 
 	// parse status & spec, create initial status object on first call
-	specAnnoData, statusAnnoData, err := r.parseAnnotations(node)
```
beekhof: `Reconcile()` is getting very large, time to break it up into some smaller functions

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: renamed to LeaseStatusActiveOwnedByDifferentHolder  ; occurs if *lease.Spec.HolderIdentity of an existing lease is different from name of the current node.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: two lines earlier I am setting an error state. Please suggest what can be done besides this.  

                        statusAnnoData.status = NodeStateLeaseNotOwnedByNode                                                                                                                  
                        r.setStatusAnnotation(node, statusAnnoData)

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -1,36 +1,105 @@
...
+	//NodeStateFailedToCordon      NodeMaintenanceStatusType =  "Error,FailedToLeaseCordon"
+	//NodeStateFailedToEvictDelete NodeMaintenanceStatusType = "Error,FailedToEvictDelete"
+)
+
+// parse this out of NodeMaintenanceSpecAnnotation annotation
+type ReconcileNodeMaintenanceSpecInfo struct {
+	// duration of maintenance window in seconds.
+	requestedMaintenenceWindowSeconds int32 `json:"requestedMaintenenceWindowSeconds"`
+	// allow extension of lease when about to expired?
+	enableMaintWindowExtensions bool `json:"enableMaintWindowExtensions"`
```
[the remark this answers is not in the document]
MoserMichael: will delete the enaleMaintenanceWindowExtension field. I am not entirely sure Are you sure that you will not like to add additional input parameters ?

pkg/controller/nodemaintenance/taint.go
MoserMichael: brought it back.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -39,284 +108,413 @@ func Add(mgr manager.Manager) error {
...
+	}
+
+	return reconciler, nil
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
-	c, err := controller.New("nodemaintenance-controller", mgr, controller.Options{Reconciler: r})
+	c, err := controller.New("NodeMaintenance-controller", mgr, controller.Options{Reconciler: r})
```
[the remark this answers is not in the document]
MoserMichael: fixed.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
MoserMichael: changed to debug

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: fixed.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: deleted it.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
MoserMichael: I have cut lease extension. Now are you sure we won't be adding additional input fields to the spec annotation?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: good question - I didn't know of a simple way to maintain additional state with the reconcile request. Do you have an idea? (actually the PR you sited creates the drainer again and again)

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: not yet tested. I don't know how to design an experiment that could prove this, as this is a very transient state. I guess that I should look at what kubelet is doing with this field in order to find out.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: broke it up into shorter pieces.

pkg/controller/add_nodemaintenance.go
MoserMichael: renamed back into nodemaintenance

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: right now I am not checkingif the lease period has been modified vs that of the lease . Will have to add this check on every reconcile call.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
beekhof: If we need 1 or 2 extra fields, we can always add additional annotations.
If we find we need more than that, then we should probably go back to a CRD

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: Perhaps add it to `r` ?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +133,487 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+					// TODO: is that the right thing? or should it rather report error.
+					return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime * time.Second}, nil
+				}
+			} else {
+				return reconcile.Result{}, err
+			}
+			statusAnnoData.Status = NodeStateEvicting
+			_, err := r.setStatusAnnotation(node, statusAnnoData)
+			if err != nil {
+				return reconcile.Result{}, err
```
beekhof: This demonstrates the fragility of relying on `statusAnnoData.Status`

If ``setStatusAnnotation()`` fails, it does not mean that we are in state `NodeStateCordon`, nor that cordon needs to be reattempted, nor that eviction should be delayed - but using `statusAnnoData.Status` in this way asserts that all these things are true anyway.

It would be better to check the actual condition rather than a flag that may be out of sync with reality.
In this specific case:  
````
nodeCorodoned = ...
if !nodeCordoned  { r.runCordonOrUncordon() ... }
````

I'm not saying to delete `statusAnnoData.Status`, but let it just be for providing information to admins rather than having the controller make decisions based on its value.

File: pkg/controller/nodemaintenance/nodemaintenance_controller.go [hunk 16]
```diff
@@ -55,268 +133,487 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			reqLogger.Infof("NodeMaintenance Object: %s Deleted ", request.NamespacedName)
-			return reconcile.Result{}, nil
+	switch statusAnnoData.Status {
+	case NodeStateInit:
```
beekhof: This state seems unnecessary

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +133,487 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			_, err := r.setStatusAnnotation(node, statusAnnoData)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+		fallthrough
+
+	case NodeStateEvicting:
+		{
+			if drainer, _, err := r.initDrainer(); err == nil {
```
beekhof: As above, do this based on whether there are pods to drain, not the value of ``statusAnnoData.Status``

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +133,487 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+
+			statusAnnoData.Status = NodeStateMaintenanceMode
+			_, err := r.setStatusAnnotation(node, statusAnnoData)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+		fallthrough
+
+	case NodeStateMaintenanceMode:
```
beekhof: Unnecessary state

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: added that too.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +133,487 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-			// Stop node maintenance - uncordon and remove live migration taint from the node.
-			if err := r.stopNodeMaintenance(instance.Spec.NodeName); err != nil {
-				if errors.IsNotFound(err) == false {
-					return r.reconcileAndError(instance, err)
+		fallthrough
+
+	case NodeStateUnEvicting:
+		{
+			// cancelling evictions
+			if drainer, _, err := r.initDrainer(); err == nil {
```
beekhof: Similar to ``processMaintModeOn``, do things based on the state of the system not  ``statusAnnoData.Status``

For example, in this case, "am i about to loose the lease and are there pods being evicted?"   

This might imply that it is necessary to add some amount of extra time when creating the lease so that the controller has a window where it has the authority to undo the work of ``processMaintModeOn``

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: 1. Create a replica set with as many copies as worker nodes, and use a label selector to require the label "evictionTest"
1. Create a PDB for the replica set's pods with maxUnavailable=1
1. Add "evictionTest" to all workers except one

Now you have a set of pods that cannot be evicted (due to the PDB)
Next:

1. Run the eviction loop on a node where the replica set is active
1. See that the pod is not evicted, record the pod's yaml
1. Run `cancelEviction()` and see what changes happen to the pod's yaml

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
MoserMichael: switched the input annotation value from json to integer value.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: i am not sure what this is referring to: had to squash all previous commits because DCO signoff wouldn't pass.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: You need it if the lease is initially owned by some other holder, in this case the request is requeued (to distinguish the case from a real error) (line #196 right now)

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: How do you indicate that transition to maintenance mode has completed?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +133,487 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+}
+
+// Reconcile monitors Nodes and creates the MachineRemediation object when the node has reboot annotation.
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+
+	r.reqLogger = log.WithFields(log.Fields{"Request.Namespace": request.Namespace, "Request.Name": request.Name})
+	r.reqLogger.Debug("Reconciling ReconcileNodeMaintenance")
```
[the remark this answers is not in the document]
MoserMichael: deleted.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: r is ReconcileNodeMaintenance - we have one such object per control loop; therefore we would need to add  a map from Request object to some structure that holds the drainer and delete that one when the request is completed. Do you approve of this course of action?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: Ok, I would keep the state of the request somewhere. That leads us to the same problem as with keeping the drainer per reconcile request:  
I can't just keep it in the ReconcileNodeMaintenance structure - we have one such object per control loop; therefore we would need to add a map from Request object to some structure that holds the status (and drainer) and delete that one when the request is completed. Do you approve of this course of action?

MoserMichael: I can't extend the reconcile.Request structure either - it is created inernally and I can't hook the creation of the request.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +133,487 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-			return reconcile.Result{}, nil
+	switch statusAnnoData.Status {
+	case NodeStateInit:
+		{
+			statusAnnoData.Status = NodeStateCordon
+			if _, err := r.setStatusAnnotation(node, statusAnnoData); err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+		fallthrough
```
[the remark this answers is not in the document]
MoserMichael: fixed.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: You already did that in the previous state.
Nothing happens here

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: > Ok, I would have keep the state of the request somewhere. 

Yes, keep it somewhere, in the same place it is now BUT only to provide information to the admin, NOT to use as part of the state machine logic.

> That leads us to the same problem as with keeping the drainer per reconcile request:
> I can't just keep the state of the reconcile request in the ReconcileNodeMaintenance structure - we have one such object per control loop; therefore we would need to add a map from Request object to some structure that holds the status (and drainer) and delete that one when the request is completed. Do you approve of this course of action?

No. See above

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		return reconcile.Result{}, err
 	}
-	return nil
+
+	if specAnnoData != nil { // maint. annotation exists. user requests to transition toware
+		if ret, res, err := r.handleLease(specAnnoData, statusAnnoData, node); ret {
+			return *res, err
+		}
+
+		res, err := r.processMaintModeOn(statusAnnoData, node)
```
beekhof: Nit: probably just do `return r.processMaintModeOn(statusAnnoData, node)`

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		return res, err
+
+	} else if statusAnnoData != nil { // spec deleted but annotation exists -> can be request to exit maintenance mode.
+
+		res, err := r.processMaintModeOff(statusAnnoData, node)
+		return res, err
+	}
+
+	// nothin to do on the node, no relevent annotations have been set
```
beekhof: What if the lease exists and is owned by us?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-	//Ignore DaemonSet-managed pods.
-	//This is required because every node running a VirtualMachineInstance will also be running our helper DaemonSet called virt-handler.
-	//This flag indicates that it is safe to proceed with the eviction and to just ignore DaemonSets.
-	r.drainer.IgnoreAllDaemonSets = true
+			statusAnnoData.Status = NodeStateLeaseNotOwnedByNode
+			r.setStatusAnnotation(node, statusAnnoData)
 
-	//Period of time in seconds given to each pod to terminate gracefully. If negative, the default value specified in the pod will be used.
-	r.drainer.GracePeriodSeconds = -1
+			err := fmt.Errorf("lease got acquired by someone while request was active. node: %s", nodeName)
```
beekhof: This is not an error.
You would end up here if someone already had the lease prior to the maintenance window being requested

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+func (r *ReconcileNodeMaintenance) handleLease(specAnnoData *ReconcileNodeMaintenanceSpecInfo, statusAnnoData *ReconcileNodeMaintenanceStatusInfo, node *corev1.Node) (bool, *reconcile.Result, error) {
+
+	nodeName := node.ObjectMeta.Name
+
+	// handling of lease as a prerequisite to state transitions.
+	leaseStatus, err, lease := r.doesLeaseExist(r.client, nodeName)
+	switch leaseStatus {
+	case LeaseStatusExpired:
+		{
+			if err := r.client.Delete(context.TODO(), lease); err != nil {
```
beekhof: Why do we need to delete it instead of updating the holder and expiration?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-type ReconcileNodeMaintenance struct {
-	// This client, initialized using mgr.Client() above, is a split client
-	// that reads objects from the cache and writes to the apiserver
-	client      client.Client
-	scheme      *runtime.Scheme
-	drainer     *drain.Helper
-	podInformer cache.SharedInformer
+func (r *ReconcileNodeMaintenance) processMaintModeOn(statusAnnoData *ReconcileNodeMaintenanceStatusInfo, node *corev1.Node) (reconcile.Result, error) {
+
+	drainer, _, err := r.initDrainer()
```
beekhof: Is there something about `drainer` that makes it unsuitable to be cached?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			reqLogger.Infof("NodeMaintenance Object: %s Deleted ", request.NamespacedName)
-			return reconcile.Result{}, nil
-		}
-		// Error reading the object - requeue the request.
-		reqLogger.Info("Error reading the request object, requeuing.")
-		return reconcile.Result{}, err
+		return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime * time.Second}, nil
```
beekhof: Don't we want to return the error too?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -1,36 +1,104 @@
...
+	LeaseStatusActiveOwnedByMe              = 1  // there is a lease active right now
+	LeaseStatusActiveOwnedByDifferentHolder = 3
+	LeaseStatusExpired                      = 4
+	LeaseStatusFail                         = 5
+)
+
+const (
+	MinMaintenenceWindowSeconds = 300 // minimum maintenance window size.a
+	EvictionTimeSlice           = 20  // max. time in seconds that eviction may block.
+	RequeuDrainingWaitTime      = 10  // time to reqeueu draining if not completed.
```
beekhof:
```suggestion
	RequeuDrainingWaitTime      = 10 * time.Second  // time to requeue draining if not completed.
```

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -1,36 +1,104 @@
...
+	LeaseStatusNotFound                     = -1 // currently not active lease
+	LeaseStatusActiveOwnedByMe              = 1  // there is a lease active right now
+	LeaseStatusActiveOwnedByDifferentHolder = 3
+	LeaseStatusExpired                      = 4
+	LeaseStatusFail                         = 5
+)
+
+const (
+	MinMaintenenceWindowSeconds = 300 // minimum maintenance window size.a
+	EvictionTimeSlice           = 20  // max. time in seconds that eviction may block.
```
beekhof:
```suggestion
	EvictionTimeSlice           = 20 * time.Second // max. time in seconds that eviction may block.
```

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -1,36 +1,104 @@
...
+	RequeuDrainingWaitTime      = 10  // time to reqeueu draining if not completed.
+	LeaseGracePeriod            = 30  // seconds before deadline when we might need to request additional lease time.
```
beekhof: Unused?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-				if errors.IsNotFound(err) == false {
-					return r.reconcileAndError(instance, err)
-				}
-			}
-			// Remove our finalizer from the list and update it.
-			instance.ObjectMeta.Finalizers = RemoveString(instance.ObjectMeta.Finalizers, kubevirtv1alpha1.NodeMaintenanceFinalizer)
-			if err := r.client.Update(context.Background(), instance); err != nil {
-				return r.reconcileAndError(instance, err)
-			}
+	if err := r.cancelEviction(drainer, node, statusAnnoData); err != nil {
```
beekhof: What if another controller starts evicting Pods after the maintenance window ends?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime * time.Second}, nil
+	}
+
+	if err := r.runCordonOrUncordon(drainer,  node, statusAnnoData, false); err != nil {
```
beekhof: What if another controller cordon's the Node after the maintenance window ends?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: Oh, I see, you're differentiating between the two cases based on `statusAnnoData.Status`
Probably reasonable, but maybe we should go back to a "waiting for the lease" state.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +131,477 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime * time.Second}, nil
+	}
+
+	leaseStatus, _, lease := r.doesLeaseExist(r.client, nodeName)
+	if leaseStatus == LeaseStatusActiveOwnedByMe {
+		if err := r.client.Delete(context.TODO(), lease); err != nil {
+			r.reqLogger.Errorf("Failed to create lease. node %s error: %v", nodeName, err)
```
beekhof:
```suggestion
			r.reqLogger.Errorf("Failed to delete lease. node %s error: %v", nodeName, err)
```

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: Are you thinking that there can be multiple calls to `Reconcile()` at the same time?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: Did you consider the second part?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: depending on the mode:  if lifecycle.openshift.io/maintenance exists

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: fixed

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: in file sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: fixed.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: I changed it to update for this case, instead of delete/create

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: in this project we have MaxConcurrentReconciles = 1 (as configuration didn't set this value), so it should be ok, until someone changes that.

MoserMichael: fixed

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: should i report an error  if an cordoning is already set when the user requests us to enter maintenance mode?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: should i report an error if an cordoning is already set when the user requests us to enter maintenance mode?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: What other piece of information could you use to decide whether to call ``cancelEviction()``?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: What other piece of information could you use to decide whether to call ``runCordonOrUncordon()`` here?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -186,16 +186,17 @@ func (r *ReconcileNodeMaintenance) handleLease(specAnnoData *ReconcileNodeMainte
 	switch leaseStatus {
 	case LeaseStatusExpired:
 		{
-			if err := r.client.Delete(context.TODO(), lease); err != nil {
-				r.reqLogger.Infof("failed to delete expired lease. node: %s err: %v\n", nodeName, err)
-
+			if err := r.createOrUpdateLease(node, specAnnoData.RequestedMaintenanceWindowSeconds, lease); err != nil {
```
beekhof: ``lease`` is always ``nil`` here.
Seems like this case and ``LeaseStatusNotFound`` could be handled together.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: There is a third case: the lease exists, is owned by us, but has expired (or is about to)

beekhof:
Tests will need to pass before we can consider merging this

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -1,36 +1,105 @@
...
+	NodeStateMaintenanceModeOff NodeMaintenanceStatusType = "maintenance-mode-off"
+
+	NodeStateFailedToGetLease    NodeMaintenanceStatusType = "Error,FailedToGetLease"
+	NodeStateFailedToCreateLease NodeMaintenanceStatusType = "Error,FailedToCreateLease"
+	NodeStateFailedToUpdateLease NodeMaintenanceStatusType = "Error,FailedToExtendLease"
+	NodeStateLeaseNotOwnedByNode NodeMaintenanceStatusType = "Error,LeaseNotOwnedByNode"
+)
+
+// parse this out of NodeMaintenanceSpecAnnotation annotation
+type ReconcileNodeMaintenanceSpecInfo struct {
```
beekhof: Arguably this type is  not needed and the variable can simply be an ``int32``

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +136,511 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		return reconcile.Result{}, err
+	}
 
-	// Watch for changes to primary resource NodeMaintenance
-	err = c.Watch(src, &handler.EnqueueRequestForObject{}, pred)
+	// parse status & spec, create initial status object on first call
+	specAnnoData, statusAnnoData, node, err := r.parseAnnotations(node)
 	if err != nil {
-		return err
+		r.reqLogger.Errorf("invalid value of annotation. name %s error %v", request.Name, err)
```
beekhof: ``request.Name`` is logged with every call.
Logging the annotation name here would be worthwhile though.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +136,511 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+
+// Reconcile monitors Nodes and creates the MachineRemediation object when the node has reboot annotation.
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 
-	// Create a source for watching noe maintenance events.
-	src := &source.Kind{Type: &kubevirtv1alpha1.NodeMaintenance{}}
+	r.reqLogger = log.WithFields(log.Fields{"Request.Namespace": request.Namespace, "Request.Name": request.Name})
```
beekhof: Do you have an example of the contents of ``request.Name`` and ``request.Namespace``?
I would expect ``request.Namespace`` to contain no useful information since all nodes are in the default namespace.
Does ``request.Name`` match the Node name?  If not, the later would be arguably more useful to know.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: bump

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: Does not seem to be reflected in the PR

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +136,511 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+
+	return false, nil, nil
 }
 
-var _ reconcile.Reconciler = &ReconcileNodeMaintenance{}
+func (r *ReconcileNodeMaintenance) adjustLeaseTimeIfRequired(specAnnoData *ReconcileNodeMaintenanceSpecInfo, statusAnnoData *ReconcileNodeMaintenanceStatusInfo, lease *coordv1beta1.Lease) error {
+	duration := statusAnnoData.OpDeadline.Sub(statusAnnoData.OpStart)
+	durationSeconds := int64(duration.Nanoseconds()) / int64(time.Second)
+
+	diffTime := int64(specAnnoData.RequestedMaintenanceWindowSeconds) - durationSeconds
```
beekhof: Why not just compare lease.Spec.LeaseDurationSeconds with the value in the annotation?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: there is a different in updating an existing lease (lease not nil, called when LeaseStatusExpired) and creating a new one (lease is nil ; in caller LeaseStatusNotFound); isn't there?  this is what the lease ref is indicating. How can they be handled together?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: the value of the annotation is included in the error that is passed up to this level and logged with this call.

MoserMichael: changed it so that the value of the annotation will be in the error passed up by parseAnnotation and that it printed with this call.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +136,511 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			r.reqLogger.Errorf("Failed to create lease. node %s error: %v", nodeName, err)
+			return err
+		}
+	} else {
+		lease.ObjectMeta.Name = nodeName
+		lease.ObjectMeta.OwnerReferences = []metav1.OwnerReference{*owner}
+		lease.Spec.HolderIdentity = &holderIdentity
+		lease.Spec.LeaseDurationSeconds = &tmpDurationInSeconds
+		lease.Spec.AcquireTime = &metav1.MicroTime{Time: time.Now()}
+		lease.Spec.RenewTime = nil
```
beekhof: not updated anywhere

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +136,511 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		lease.Spec.LeaseTransitions = nil
```
beekhof: not updated anywhere

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: i need to reset all existing fields when updating a lease (instead of deleting/creating it anew), therefore need to set RenewTime and LeaseTransition to nil; where am I wrong?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: ``r.doesLeaseExist`` was returning `lease=nil` in both cases

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: I suggested logging the annotation _name_, not it's contents

pkg/controller/nodemaintenance/generated_mock_nodemaintenance_controller.go
```diff
@@ -1,49 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
beekhof: Did you forget to re-add this for your tests?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +194,604 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	return timeNow.After(leaseDeadline)
+}
 
-	instance.Status.Phase = kubevirtv1alpha1.MaintenanceSucceeded
-	instance.Status.PendingPods = nil
-	err = r.client.Status().Update(context.TODO(), instance)
-	if err != nil {
-		reqLogger.Errorf("Failed to update NodeMaintenance with \"Succeeded\" status. Error: %v", err)
-		return r.reconcileAndError(instance, err)
+func (r *ReconcileNodeMaintenance) doesLeaseExist() (LeaseStatus, error, *coordv1beta1.Lease) {
```
beekhof: I think it would be better to pass in the node name here

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +194,604 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			err := utilerrors.NewAggregate(errs)
+			r.reqLogger.Errorf("failed got get pod for eviction %v", err)
+			return err
 		}
-		if pendingList != nil {
-			nm.Status.PendingPods = GetPodNameList(pendingList.Pods())
+
+		if !dcheck.isExpired() && len(list.Pods()) != 0 {
+
+			r.drainer.Timeout = dcheck.DurationUntilExpiration()
```
beekhof: I don't think we want this

MoserMichael: is that better? if you set a deadline we don't want to run past it - according to your design.
[
if dcheck.isSet {
       r.drainer.Timeout = dcheck.DurationUntilExpiration()
}
](url)

pkg/controller/nodemaintenance/generated_mock_nodemaintenance_controller.go
MoserMichael: the test right now  isn't using the generated mock;

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +194,604 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	// parse status & spec, create initial status object on first call
+	err := r.parseAnnotations()
 	if err != nil {
-		return err
+		r.reqLogger.Errorf("request parsing error. error %v", err)
+		return reconcile.Result{}, err
+	}
+
+	if r.specAnnoData == nil && r.statusAnnoData == nil {
+		// nothing to do here.
```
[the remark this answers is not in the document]
MoserMichael: if there was a lease with us as the holder and no annotations to handle, should we log this and delete the lease object?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +194,604 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	r.node = &corev1.Node{}
 
-	// Watch for changes to primary resource NodeMaintenance
-	err = c.Watch(src, &handler.EnqueueRequestForObject{}, pred)
+	if err := r.client.Get(context.TODO(), request.NamespacedName, r.node); err != nil {
+		if errors.IsNotFound(err) {
+			r.reqLogger.Info("node not found")
+			return reconcile.Result{}, nil
+		}
+		r.reqLogger.Errorf("node not found. name %s error %v", request.Name, err)
```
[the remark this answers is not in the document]
MoserMichael: fixing it; just one detail: the name of the request is present in the log line because of

pkg/controller/nodemaintenance/generated_mock_nodemaintenance_controller.go
MoserMichael: the tests don't need the mock right now;  Yanir uses the mock to inject his informer trick.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +194,604 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	if leaseStatus == LeaseStatusOwnedByDifferentHolder {
+		r.reqLogger.Debugf("update lease - invalid/different owner ")
+
+		if _, err = r.createOrUpdateLease(lease, TransitionInc, NodeStateNewAcquired); err != nil {
+			return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime}, nil
+		}
+
+	}
+
+	if !isLeaseValid(lease) || isLeaseDurationZero(lease) {
```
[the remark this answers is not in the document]
MoserMichael: changed the names : now it's

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -4,33 +4,140 @@ package nodemaintenance
...
+	createClient(config *rest.Config) (kubernetes.Interface, error)
+}
+
+type ClientFactoryTest struct {
+	client kubernetes.Interface
+}
+
+func (r *ClientFactoryTest) createClient(config *rest.Config) (kubernetes.Interface, error) {
+	return r.client, nil
+}
```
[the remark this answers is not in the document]
MoserMichael: moved to test

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +194,604 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime}, nil
+		}
+
+		if err := r.cancelEviction(dcheck); err != nil {
+			return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime}, nil
+		}
+		if !dcheck.isExpired() {
+			r.createOrUpdateLease(lease, TransitionNoChange, NodeStateNone) // set lease duration to 0
+		}
+
```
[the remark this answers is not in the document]
MoserMichael: added function

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -4,33 +4,140 @@ package nodemaintenance
...
+	NodeMaintenanceSpecAnnotation   = "lifecycle.openshift.io/maintenance"
+	NodeMaintenanceStatusAnnotation = "lifecycle.openshift.io/maintenance-status"
+	LeaseHolderIdentity             = "node-maintenance"
+	LeasePaddingSeconds             = 30
+	WaitForLeasePeriod              = 10 * time.Second
+)
+
+const (
+	MinMaintenenceWindowSeconds = 300              // minimum maintenance window size. (must be bigger than LeasePaddingSeconds)
+	EvictionTimeSlice           = 20 * time.Second // max. time in seconds that eviction may block.
```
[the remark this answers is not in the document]
MoserMichael: fixed

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -4,33 +4,133 @@ package nodemaintenance
...
+	RequeuDrainingWaitTime      = 10 * time.Second // time to reqeueu draining if not completed.
+)
+
+type LeaseStatus int
+
+const (
+	LeaseStatusNotFound               LeaseStatus = -1 // currently not active lease
+	LeaseStatusOwnedByMe              LeaseStatus = 1  // there is a lease active right now
+	LeaseStatusOwnedByDifferentHolder LeaseStatus = 3
+	LeaseStatusFail                   LeaseStatus = 6
```
beekhof: This is an odd sequence of values

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
beekhof: We should never delete the lease.
If the lease is currently valid, we need to uncordon, cancel drains, and "release" the lease (set  ``LeaseDurationSeconds`` to ``0``).

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +187,629 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			r.reqLogger.Errorf("failed got get pod for eviction %v", err)
 			return err
 		}
-		nm.Status.TotalPods = len(podlist.Items)
-		err = r.client.Status().Update(context.TODO(), nm)
+
+		if !dcheck.isExpired() && len(list.Pods()) != 0 {
+
+			if dcheck.isSet && r.drainer.Timeout > dcheck.DurationUntilExpiration() {
+				r.drainer.Timeout = dcheck.DurationUntilExpiration()
```
beekhof: If this is our first time through the `Reconcile()` loop for a 3h maintenance window, the line above tells the system to wait up to the entire 3 hour period before processing any other maintenance requests.

The most we should wait for any eviction, deletion, or cordon request is:
``EvictionTimeSlice           = 10 * time.Second // max. time in seconds that eviction may block.``

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -55,268 +194,604 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		r.reqLogger.Debugf("update lease - lease duration change requested")
+
+		if _, err = r.createOrUpdateLease(lease, TransitionNoChange, NodeStateUpdated); err != nil {
+			return reconcile.Result{Requeue: true, RequeueAfter: RequeuDrainingWaitTime}, nil
 		}
+
+	}
+
+	if isLeaseExpired(lease, false) {
+		r.reqLogger.Debugf("update lease - lease expired")
```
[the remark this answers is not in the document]
beekhof: Comment does not seem to have moved with the code as it changed.  Applies to ``NodeStateEnded`` case

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -4,33 +4,133 @@ package nodemaintenance
...
+var _ reconcile.Reconciler = &ReconcileNodeMaintenance{}
+
+const (
+	NodeMaintenanceSpecAnnotation   = "lifecycle.openshift.io/maintenance"
+	NodeMaintenanceStatusAnnotation = "lifecycle.openshift.io/maintenance-status"
+	LeaseHolderIdentity             = "node-maintenance"
+	LeasePaddingSeconds             = 30
+	WaitForLeasePeriod              = 10 * time.Second
+)
+
```
yanirq: If any of these constants are subjected to user config then i would suggest adding them as a field under the maintenance CRD (can be done in a different PR)

yanirq:
@MoserMichael  Can you add a full description to this PR explaining all the changes/what was added?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
yanirq: the namespace you will get here is probably the namespace of the operator itself , request name will most likely have the node maintenance CR spec.name i believe (not sure it will hold the node name)

yanirq:
I see that finalizer was removed, what will happen now if we delete the maintenance CR ?  

I'm still worried about the logic inside the controller that might take it away from being idempotent. Are we relaying on state to perform operations inside the loop and change a state accordingly ? reconcile should be eventually a stateless operation.

MoserMichael:
> @MoserMichael Can you add a full description to this PR explaining all the changes/what was added?

- the operator is not based on CRD; it is based on node annotations: Andrew says that's because the openshift people didn't want to add another CRD to the system.

   to start transition to maintenance mode on node add  lifecycle.openshift.io/maintenance annotation where value is the number of seconds it has to be in maintenance mode. 

 
 the current state of the show is keps in node annotation: lifecycle.openshift.io/maintenance-status
   - the value "active" indicates that you are in maintenance mode.
   - the value "ended" means you finished with the maintenance mode.

to end maintenance mode delete the lifecycle.openshift.io/maintenance
- use upstream drain library

- utilise lease objects; k8s.io/api/coordination/v1beta1/Lease - so that for the transition and duration of the maintenance mode there is an active lease with holderIdentity  equal to node-maintenance-operator; that's in addition to cordoning the node and setting the  taints.

- Reconcile is non blocking/asynchronous manner

yanirq:
@MoserMichael  @beekhof  For all the changes described in this PR I would have written a new controller/deamonset/operator ;)   plus this is a huge change of direction, are we sure we want to get rid of the CRDs in this repository ?  if so, im not sure this current operator-sdk framework is clean and suitable for these changes...

beekhof:
@MoserMichael I thought you said all the tests were passing?

beekhof:
> are we sure we want to get rid of the CRDs in this repository ? 

At the point Michael started, we still thought we were going to be delivering this through OpenShift somehow, and they were very negative about having a new CRD.  Since then I've gotten agreement that as long as we're co-ordinating via node leases, then how the NMO is triggered is an implementation detail.  Plus we now know that we'll need the CRD for OLM's dependency model (so that CNV can automatically depend on it instead of having the HCO deploy it... one day).

The original request was to add node leases to the existing controller, but Michael wanted to tackle everything at once.  So I'm giving him the opportunity to see that through and then we'll revisit the CRD/annotation question.

I don't think there is much to review until the tests are passing.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -4,33 +4,147 @@ package nodemaintenance
...
+	clientset kubernetes.Interface
+	drainer   *drain.Helper
+
+	// used only for the duration of one reconcile call. transient value.
+	// That's ok as there is one go routine that handles reconciles (here)
+	reqLogger      *log.Entry
+	node           *corev1.Node
+	specAnnoData   *ReconcileNodeMaintenanceSpecInfo
+	statusAnnoData *ReconcileNodeMaintenanceStatusInfo
```
beekhof: Data specific to an individual request is not appropriate to be stored in this structure.

beekhof:
I opened this PR to be better able to give feedback during development.
Michael, can you create a new one in your own name once you are ready for review.

PR #52, opened 2020-02-19
Do not merge: consume upstream drain

beekhof:


yanirq:
You might be able to remove now the podinformer that was used by the custom drain lib we had before (https://github.com/kubevirt/node-maintenance-operator/blob/d3b837cbd73fd8bd264826751c5691090b80ef2c/pkg/controller/nodemaintenance/nodemaintenance_controller.go#L156)  
This means you will need to remove the interface (reconciler handler) and the mocks for it in the testing framework

PR #59, opened 2020-03-09
Prevent NMO from being deployed on master nodes

omeryahud:
This PR proposes to prevent NMO pods from being deployed on master nodes, as it is a bad practice to deploy 3rd party apps on master nodes in OCP.

### DIFF
--- a/test/e2e/nodemaintenance_test.go
+++ b/test/e2e/nodemaintenance_test.go
@@ -128,15 +128,26 @@ func nodeMaintenanceTest(t *testing.T, f *framework.Framework, ctx *framework.Te
 		return fmt.Errorf("could not get namespace: %v", err)
 	}
 
+	// Count worker nodes to spread the test deployment on all of them
+	computeNodesNumber := countWorkerNodes(t, f)
+
+	if computeNodesNumber < 2 {
+		t.Fatal("not enough worker nodes in the cluster, expecting at least 2")
+	}
+
+	testDeploymentReplicas = int32(computeNodesNumber)
+
+	// Create a test deployment with pod anti affinity to have a pod on each node
 	err = createSimpleDeployment(t, f, ctx, namespace, testDeploymentReplicas)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	nodeName, err := getCurrentDeploymentHostName(t, f)
-	if err != nil {
-		t.Fatal(err)
-	}
+	// Drain the node the operator is deployed on, in order to test
+	// if it will be recovered and complete the drain operation from a different node
+	// note: this node will have a test deployment pod because of its replica count
+	// and anti pod affinity
+	nodeName := getOperatorHostname(t, f)
 
 	t.Logf("Putting node %s into maintanance", nodeName)
 
@@ -156,7 +167,12 @@ func nodeMaintenanceTest(t *testing.T, f *framework.Framework, ctx *framework.Te
 	}
 
 	// Create node maintenance CR
-	err = f.Client.Create(goctx.TODO(), nodeMaintenance, &framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
+	err = f.Client.Create(goctx.TODO(), nodeMaintenance, &framework.CleanupOptions{
+		TestContext: ctx,
+		Timeout: cleanupTimeout,
+		RetryInterval: cleanupRetryInterval,
+	})
+
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


yanirq:
This calls for a test plane and a CI e2e test to see what happens when a node with NMO is drained

dankenigsberg:
@omeryahud why do you need this PR? Please include your motivation in every commit message and every description of every PR.

beekhof:
We need to know the context for this change before we can consider it.
Doesn't make much sense at first glance.

omeryahud:
@yanirq Yes you are correct, I just opened this PR to see if any of the test lanes would fail.
@dankenigsberg @beekhof The motivation for this PR is to prevent NMO from being deployed on master nodes, because it is considered a bad practice to deploy 3rd party apps on master nodes in OCP.
I opened this PR just to see if any of the test lanes would fail, I suppose I should have declared it as Work in Progress.

beekhof:
I would argue that it shouldn't be a 3rd party operator, but that's a different conversation.

IMHO the complexity of the NMO potentially evacuating itself trumps the "no masters" best practice in this case.  It would make the admin experience demonstrably worse.

omeryahud:
If I'm not wrong, the e2e tests in NMO test exactly that, that NMO manages to drain the node it's deployed on. 
In any case, I would argue that the place for NMO in OCP is on an infra node.
With regards to k8s, in order to drain a node, you must have at least one other node that is schedulable, so in case NMO is deployed on the drained node, it would just redeploy on the other node and continue draining the target node, this is the process I am seeing with manual testing atm.

omeryahud:
@yanirq @beekhof @ksimon1 could any you guys take a look please?

test/e2e/nodemaintenance_test.go
```diff
@@ -354,3 +362,74 @@ func checkFailureStatus(t *testing.T, f *framework.Framework) {
...
+	for _, pod := range pods.Items {
+		if !sliceContainsString(hostnames, pod.Spec.NodeName) {
+			hostnames = append(hostnames, pod.Spec.NodeName)
+		}
+	}
+
+	return hostnames
+}
+
+func sliceContainsString(slice []string, str string) bool {
```
yanirq: maybe create a utils folder/class for that under test folder and place this function there , as in:
https://github.com/kubevirt/node-maintenance-operator/blob/master/pkg/controller/nodemaintenance/utils.go

test/e2e/nodemaintenance_test.go
```diff
@@ -354,3 +362,74 @@ func checkFailureStatus(t *testing.T, f *framework.Framework) {
...
+
+	if len(pods.Items) < 1 {
+		t.Fatal("there are no operator pods")
+	}
+
+	return pods.Items[0].Spec.NodeName
+}
+
+func countWorkerNodes(t *testing.T, f *framework.Framework) int {
+	nodes := corev1.NodeList{}
```
yanirq: another suggestion to get all nodes with a label:
```
// GetByRole returns all nodes with the specified role
func GetByRole(c client.Client, role string) ([]corev1.Node, error) {
	selector, err := labels.Parse(fmt.Sprintf("%s/%s=", testutils.LabelRole, role))
	if err != nil {
		return nil, err
	}

	nodes := &corev1.NodeList{}
	if err := c.List(context.TODO(), nodes, &client.ListOptions{LabelSelector: selector}); err != nil {
		return nil, err
	}
	return nodes.Items, nil
}
```

Can move to a util class

omeryahud: Problem is, I don't think worker nodes have roles applied (or not all of the time):
```
$ kubectl get nodes
NAME     STATUS   ROLES    AGE   VERSION
node01   Ready    master   20h   v1.16.2
node02   Ready    <none>   20h   v1.16.2
node03   Ready    <none>   20h   v1.16.2
```

test/e2e/nodemaintenance_test.go
omeryahud: Yea, sounds good, will do

test/e2e/nodemaintenance_test.go
yanirq: right.  well if some of the code lines here are more elegant feel free to use them anyway

omeryahud:
@beekhof @ksimon1 could you guys take a look please?

beekhof:
I don't doubt that it's a good implementation, but I'm not seeing that there is a compelling reason for this change.

Is there anything you can point that illustrates this operator running on a master is causing a problem?  

IMHO, it is a deficiency that the NMO is not part of OCP and that masters are exactly where the NMO _should_ be running.

omeryahud:
> Is there anything you can point that illustrates this operator running on a master is causing a problem?

Nothing other than OCP best practices.

> IMHO, it is a deficiency that the NMO is not part of OCP and that masters are exactly where the NMO _should_ be running.

Who should be involved and come up with the decision to make it part of OCP?

fabiand:
I suppose the easy answer is: NMO is not a master.
Technically, and please keep me honest, it should not matter if NMO is running on a worker or master.

A servicability reason for not running NMO on masters is simply to make and keep the resource consumption of masters predictable by leaving those nodes to core master components.

Certain K8s distributions like OCP also have concepts like infra nodes, which are hosting operators, that is actually where I'd imagine NMO to live.
Or ideally - on the long run - become a part of the core of Kubernetes ... let the dreams live on.

fabiand:
ping?

slintes:
Hey @fabiand, small world :wink: 

I would be happy to drive this PR forward, but from the comments I'm not sure if there is an agreement on how to go on.
My take: 
> A servicability reason for not running NMO on masters is simply to make and keep the resource consumption of masters predictable by leaving those nodes to core master components.

Because of this I would also say that NMO should not run on masters.
But we should make sure that NMO doesn't take itself down by adding a `PodDisruptionBudget` to its deployment. Does that make sense? @beekhof wdyt?

beekhof:
What does the patch do if there are only masters?
Is a PDB overkill for a single pod?

beekhof:
What I would like to avoid is the UX nightmare that the console has when it runs out of workers to run on.

If we can make it prefer workers, but run on masters when there is no other choice (either there were never workers in the cluster, or they failed), that is something I could get onboard with.

fabiand:
Hello (waving from my continent to the next).

I understand (I think I do) @beekhof concerns, however, if there are no workers, then we will have many issues with other worklaods and operators as well.
I'd recommend to take this to some OpenShift discussion place to hear their opinions.

because technically we can run it anywhere, I'm just looking if we can make the cluster behavior over all more predictable.

beekhof:
Looks like David has a thread called "how to influence operator pod location when installing via olm subscription?", but it doesn't seem like there are good answers.

beekhof:
> What does the patch do if there are only masters?
> Is a PDB overkill for a single pod?

@omeryahud  Do we have an answer for this?

slintes:
The discussion about being installed on master node revived, and we decided to follow kubevirt and friends by removing the affinity, because it also conflicts with a potential node selector on the OLM subscription. Handling this in #140 instead of trying to revive this PR though, sorry.

/close

PR #61, opened 2020-03-10
align CI scripts with hyperconverged cluster operator.

MoserMichael:

Align CI scripts with hyperconverged cluster operator. That was the only project that I could find with a working CI for openshift 4.3 at the time of writing.

**Special notes for your reviewer:**

**Release note:**
```
NONE
```


ovirt-infra:

beekhof:
I would have expected the old scripts to be removed as well

beekhof:
@yanirq what do you think?

This is a pre-req for updating the CI jobs and adding use of node leases

beekhof:
Need to complete the release note process

yanirq:
> @yanirq what do you think?
> 
> This is a pre-req for updating the CI jobs and adding use of node leases

I wonder if we really need to duplicate all the code here for CI lanes. 
Maybe we can utilize [KubeVirt CI](https://github.com/kubevirt/kubevirtci/tree/master/cluster-up/cluster ) here in some way instead of duplications ? 
Adding some external possible advice here :)   @cynepco3hahue  @gbenhaim  any opinions on that matter ?

beekhof:
@yanirq I expect that the duplication is necessary since even the HCO is doing so.

However, perhpas `git subtree` and some Makefile helpers would make the process of keeping it in sync less painful.

MoserMichael:
I will try to work on a new PR, the proposed changes are quite profound. I am not sure that the upstream kubevirt ci works with openshift 4.x right now, but I will try. In any event we have this PR as a fallback option.

yanirq:
You could use git-subtrees but if something breaks it could be a pitfall . just thinking out loud. 
this can be re-factored later on. just trying to think of an elegant solution

MoserMichael:
> You could use git-subtrees but if something breaks it could be a pitfall . just thinking out loud.
> this can be re-factored later on. just trying to think of an elegant solution

why not just accept this PR as is?

MoserMichael:
https://github.com/kubevirt/node-maintenance-operator/pull/67/ please review the PR that uses kubevirtCI.

beekhof:
If #67 replaces this PR, please close.

beekhof:
If #67 replaces this PR, please close.

PR #67, opened 2020-03-22
Consume cluster-up from kubevirtci using git subtree

MoserMichael:
**What this PR does / why we need it:**

try to use CI scripts from kubevirt.
pull kubevirt ci via git subtree command.

**Special notes for your reviewer:**

**Release note:**
```
NONE
```

PR #68, opened 2020-03-29
fix openapi validation error for CRD of node maintenance operator.

MoserMichael:
attempt to fix an error while validating the openapi schema. starting point is branch with the new CI.
I had to build the PR on top of PR #67 (updated CI) so as to be able to check it with openshift 4.3. The last two commits of the PR are a check for openAPI validation errors in the e2e tests, and the fix of the CRD


test/e2e/nodemaintenance_test.go
```diff
@@ -30,6 +32,30 @@ var (
 	podLabel             = map[string]string{"test": "drain"}
 )
 
+const (
+	scriptShowCRD = `./kubevirtci/cluster-up/kubectl.sh get -o yaml crd nodemaintenances.kubevirt.io`
```
beekhof: This is `Go`, is there not a programatic way to obtain the yaml of a CRD from the server?

test/e2e/nodemaintenance_test.go
```diff
@@ -74,6 +100,8 @@ func ClusterTest(t *testing.T) {
 }
 
 func nodeMaintenanceTest(t *testing.T, f *framework.Framework, ctx *framework.TestCtx) error {
+
+
```
beekhof: No pointless whitespace changes

test/e2e/nodemaintenance_test.go
```diff
@@ -229,6 +257,7 @@ func nodeMaintenanceTest(t *testing.T, f *framework.Framework, ctx *framework.Te
 		t.Fatal(err)
 	}
 
+	checkOperatorCRD(t)
```
beekhof: Seems like something that could/should be part of a functional test rather than e2e testing

deploy/crds/nodemaintenance_crd.yaml
```diff
@@ -64,6 +64,7 @@ spec:
               format: int64
               type: integer
           type: object
+      type: object
```
beekhof: Is it necessary that `type` is the last line of an element?
Seems like it might be better after line 16, where it's clearer which earlier elements it refers to.

MoserMichael: tried to move it, but build failed with the following error:

```
./build/make-csv-generator.sh

time="2020-03-31T01:26:09Z" level=info msg="Generating CSV manifest version 9999.9999.9999"

Error: catalog scaffold failed: (deploy/crds/nodemaintenance_crd.yaml: error converting YAML to JSON: yaml: line 19: mapping values are not allowed in this context)
```

test/e2e/nodemaintenance_test.go
MoserMichael: can't put it in the unit test in pkg/controller/nodemaintenance/nodemaintenance_controller_test.go ; validation of schema is done by apiserver, whereas that test runs on fake clients. We need a deployed system for the test.

test/e2e/nodemaintenance_test.go
MoserMichael: removed whitespace changes.

test/e2e/nodemaintenance_test.go
MoserMichael: kubectl seems to be running this function in order to do the kubectl get command;
what i did't find so far is how they handle the -o json option.

repo: https://github.com/kubernetes/kubernetes.git
file: ./staging/src/k8s.io/kubectl/pkg/cmd/get/get.go                                                                                                                             
```
// Run performs the get operation.
// TODO: remove the need to pass these arguments, like other commands.
func (o *GetOptions) Run(f cmdutil.Factory, cmd *cobra.Command, args []string) error {
	if len(o.Raw) > 0 {
		restClient, err := f.RESTClient()
		if err != nil {
			return err
		}
		return rawhttp.RawGet(restClient, o.IOStreams, o.Raw)
	}
	if o.Watch || o.WatchOnly {
		return o.watch(f, cmd, args)
	}

	chunkSize := o.ChunkSize
	if o.Sort {
		// TODO(juanvallejo): in the future, we could have the client use chunking
		// to gather all results, then sort them all at the end to reduce server load.
		chunkSize = 0
	}

	r := f.NewBuilder().
		Unstructured().
		NamespaceParam(o.Namespace).DefaultNamespace().AllNamespaces(o.AllNamespaces).
		FilenameParam(o.ExplicitNamespace, &o.FilenameOptions).
		LabelSelectorParam(o.LabelSelector).
		FieldSelectorParam(o.FieldSelector).
		ExportParam(o.Export).
		RequestChunksOf(chunkSize).
		ResourceTypeOrNameArgs(true, args...).
		ContinueOnError().
		Latest().
		Flatten().
		TransformRequests(o.transformRequests).
		Do()


```

deploy/crds/nodemaintenance_crd.yaml
beekhof: thats an odd thing for it to enforce

test/e2e/nodemaintenance_test.go
beekhof: fyi, the canonical location is

    https://github.com/kubernetes/kubectl/blob/HEAD/pkg/cmd/get/get.go

Why not import the `get` package in the same way that we consume the `drain` package, and do what their test code does?

    https://github.com/kubernetes/kubectl/blob/HEAD/pkg/cmd/get/get_test.go#L144

MoserMichael: This one file is requiring a whole lot of dependencies - and it will be quite hard to get them working all togather.

MoserMichael:
cherry picked changes in this PR on top of master - as a result this one is superseded by #70

PR #70, opened 2020-04-05
fix openapi validation error for CRD of node maintenance operator

MoserMichael:
fix an error while validating the openapi schema. 
The last two commits of the PR are a check for openAPI validation errors in the e2e tests, and the fix of the CRD
This PR is created from #68 by cherry picking its changes of top of the new master branch.

### DIFF
--- a/hack/functest.sh
+++ b/hack/functest.sh
@@ -15,7 +15,7 @@ if [ -z "$KUBEVIRTCI_CONFIG_PATH" ]; then
     )"
 fi
 
-
+KUBECTL_CMD="${KUBEVIRTCI_PATH}/kubectl.sh"
 
 function new_test() {
     name=$1
@@ -42,4 +42,31 @@ KUBE_CONFIG=$(${KUBEVIRTCI_PATH}/kubeconfig.sh)
 TEST_NAMESPACE=node-maintenance-operator go test ./test/e2e/... -root=$(pwd) -kubeconfig=${KUBE_CONFIG} -globalMan _out/nodemaintenance_crd.yaml --namespacedMan _out/namespace-init.yaml -singleNamespace
 
 echo "E2e tests passed"
+
+echo "check validation of openaAPIV3Schema"
+
+$KUBECTL_CMD create -f _out/namespace-init.yaml
+
+$KUBECTL_CMD create -f _out/nodemaintenance_crd.yaml
+
+echo "validate CRD"
+
+VALIDATE_CRD=$($KUBECTL_CMD get -o yaml crd nodemaintenances.kubevirt.io)
+if [[ $VALIDATE_CRD == "" ]]; then
+	echo "can't validate CRD, check if deployment is running"
+	exit 1
+fi
+
+set +e
+VALIDATION_ERRORS=$(echo "$VALIDATE_CRD" | grep -c "spec.validation.openAPIV3Schema")
+set -e
+
+if [[ $VALIDATION_ERRORS != "0" ]]; then
+	echo "validation of CRD failed"
+	exit 1
+fi
+
+echo "check validation of openaAPIV3Schema passed"
+
+
 
--- a/hack/functest.sh
+++ b/hack/functest.sh
@@ -15,6 +15,7 @@ if [ -z "$KUBEVIRTCI_CONFIG_PATH" ]; then
     )"
 fi
 
+
 KUBECTL_CMD="${KUBEVIRTCI_PATH}/kubectl.sh"
 
 function new_test() {
@@ -71,7 +72,7 @@ VALIDATION_ERRORS=$(echo "$VALIDATE_CRD" | grep -c "spec.validation.openAPIV3Sch
 if [[ $VALIDATION_ERRORS != "0" ]]; then
 	echo "validation of CRD failed"
 	echo "Validation errors:"
-	echo "$VALIDATE_CRD" | grep "spec.validation.openAPIV3Schema"
+	echo "$VALIDATION_ERRORS"
 	exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


beekhof:
> /ok-to-test

You're part of the kubevirt Org now, you don't need to add this flag

test/e2e/nodemaintenance_test.go
```diff
@@ -74,6 +76,28 @@ func showDeploymentStatus(t *testing.T, f *framework.Framework) {
 	t.Fatalf("operator logs: %s", str)
 }
 
+const (
+	scriptShowCRD = `./kubevirtci/cluster-up/kubectl.sh get -o yaml crd nodemaintenances.kubevirt.io`
```
beekhof, line 80: I don't like that we're using shell here, but the programatic approaches investigated so far have too many dependancies.

test/e2e/nodemaintenance_test.go
```diff
@@ -282,6 +306,7 @@ func nodeMaintenanceTest(t *testing.T, f *framework.Framework, ctx *framework.Te
 		t.Fatal(err)
 	}
 
+	checkOperatorCRD(t)
```
beekhof, line 309: I'd love to have this as a functional test rather than e2e, but it doesn't seem possible

test/e2e/nodemaintenance_test.go
MoserMichael: Removed the changes from the golang e2e tests. Now all the test are run as part of functest.sh

hack/functest.sh
```diff
@@ -43,3 +43,30 @@ TEST_NAMESPACE=node-maintenance-operator go test ./test/e2e/... -root=$(pwd) -ku
 
 echo "E2e tests passed"
 
+echo "check validation of openaAPIV3Schema"
+
+$KUBECTL_CMD create -f _out/namespace-init.yaml
+
+$KUBECTL_CMD create -f _out/nodemaintenance_crd.yaml
```
beekhof, line 50: Does the test erase these on completion?

hack/functest.sh
```diff
@@ -43,3 +43,30 @@ TEST_NAMESPACE=node-maintenance-operator go test ./test/e2e/... -root=$(pwd) -ku
...
+	echo "can't validate CRD, check if deployment is running"
+	exit 1
+fi
+
+set +e
+VALIDATION_ERRORS=$(echo "$VALIDATE_CRD" | grep -c "spec.validation.openAPIV3Schema")
+set -e
+
+if [[ $VALIDATION_ERRORS != "0" ]]; then
+	echo "validation of CRD failed"
```
beekhof, line 65: Should we display the errors?

hack/functest.sh
```diff
@@ -43,3 +43,30 @@ TEST_NAMESPACE=node-maintenance-operator go test ./test/e2e/... -root=$(pwd) -ku
...
 
+echo "check validation of openaAPIV3Schema"
+
+$KUBECTL_CMD create -f _out/namespace-init.yaml
+
+$KUBECTL_CMD create -f _out/nodemaintenance_crd.yaml
+
+echo "validate CRD"
+
+VALIDATE_CRD=$($KUBECTL_CMD get -o yaml crd nodemaintenances.kubevirt.io)
```
beekhof, line 54: I think you need the `set +e` logic here too

hack/functest.sh
MoserMichael: done

hack/functest.sh
MoserMichael: added || true in both cases (to set exit status of pipeline to success in any case)

hack/functest.sh
MoserMichael: added a trap to erase on completion.

MoserMichael:
> Might also be worth running the validation first so we find out ASAP that something is wrong with the CRD.

I would rather have this at the end - I added a trap to delete the objects on script exit, If I move this test to the start then I have to remove the trap on completion and would have to check if there are existing traps to restore.

hack/functest.sh
```diff
@@ -43,3 +43,39 @@ TEST_NAMESPACE=node-maintenance-operator go test ./test/e2e/... -root=$(pwd) -ku
...
+
+$KUBECTL_CMD create -f _out/namespace-init.yaml
+
+$KUBECTL_CMD create -f _out/nodemaintenance_crd.yaml
+
+echo "validate CRD"
+
+VALIDATE_CRD=$($KUBECTL_CMD get -o yaml crd nodemaintenances.kubevirt.io || true)
+if [[ $VALIDATE_CRD == "" ]]; then
+	echo "can't validate CRD, check if deployment is running"
```
beekhof, line 65:
```suggestion
	echo "can't validate CRD, check if the CRD is installed"
```

MoserMichael: done

hack/functest.sh
```diff
@@ -43,3 +43,39 @@ TEST_NAMESPACE=node-maintenance-operator go test ./test/e2e/... -root=$(pwd) -ku
...
+	echo "validation of CRD failed"
+	echo "Validation errors:"
+	echo "$VALIDATE_CRD" | grep "spec.validation.openAPIV3Schema"
+	exit 1
+fi
+
+echo "check validation of openaAPIV3Schema passed"
+
+
+
```
beekhof, line 81:
```suggestion
```

hack/functest.sh
beekhof: What about `go test` ?

hack/functest.sh
beekhof: Wouldn't we just want the contents of VALIDATION_ERRORS?

hack/functest.sh
MoserMichael: done

ksimon1:
@beekhof is this PR ready to be merged?

PR #73, opened 2020-04-08
move node maintenance operator to its own api group

ksimon1:
https://bugzilla.redhat.com/show_bug.cgi?id=1766522
/cc @yanirq, @tiraboschi, @beekhof 

Signed-off-by: Karel Simon <ksimon@redhat.com>

### DIFF
--- a/deploy/crds/nodemaintenance_cr.yaml
+++ b/deploy/crds/nodemaintenance_cr.yaml
@@ -1,4 +1,4 @@
-apiVersion: kubevirt.io/v1alpha1
+apiVersion: nodemaintenance.kubevirt.io/v1beta1
 kind: NodeMaintenance
 metadata:
   name: nodemaintenance-xyz
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Create the example `NodeMaintenance` CR found at `deploy/crds/nodemaintenance_cr
 ```sh
 $ cat deploy/crds/nodemaintenance_cr.yaml
 
-apiVersion: kubevirt.io/v1alpha1
+apiVersion: nodemaintenance.kubevirt.io/v1beta1
 kind: NodeMaintenance
 metadata:
   name: nodemaintenance-xyz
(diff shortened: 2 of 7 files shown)


beekhof:
Will this break the UI?

ksimon1:
@beekhof It should not break the UI. @tiraboschi  can you please confirm?

beekhof:
Iirc, the ui is enabled based on whether it finds the crd, and since this changes where the crd lives... or am i missing something?

jtomasek:
Right it will break the UI. The CRD model will need to be updated there: https://github.com/openshift/console/blob/master/frontend/packages/metal3-plugin/src/models.ts#L19

ksimon1:
PR with UI change https://github.com/openshift/console/pull/4967

deploy/crds/nodemaintenance_cr.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: kubevirt.io/v1alpha1
+apiVersion: nodemaintenance.kubevirt.io/v1alpha1
```
tiraboschi, line 1: Please notice that in order to go GA this should be at least v1beta1, maybe it's worth to address both in a single PR.

ksimon1:
@beekhof can you please review this PR?

beekhof:
Can we drop the unrelated whitespace changes to the test file, or move them into a different commit please?
Otherwise it looks good

ksimon1:
> Can we drop the unrelated whitespace changes to the test file, or move them into a different commit please?
> Otherwise it looks good

done

beekhof:
/approve

@n1r1 or @yanirq can you confirm with lgtm if you're happy?

n1r1:
PR is already merged but - does it make sense that the yaml files were changed manually to `v1beta1` but there are still many references to `v1alpha1` in the code? 

For example, the APIs are under `v1alpha1` [package](https://github.com/kubevirt/node-maintenance-operator/tree/master/pkg/apis/kubevirt/v1alpha1)

PR #76, opened 2020-05-07
Disallow master nodes to be taken into maint. mode if that violates quorum size.

MoserMichael:
Check that the node to be taken into maintenance mode is a master node; if that is the case, then  then the number of  available master nodes is not allowed to drop below the required quorum size. Fail the reconcile request with error if that is about to happen.

Unit tests are implemented as table tests.
 
This was added as a fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1826914

**Release note:**
```release-note
Check that the node to be taken into maintenance mode is a master node; if that is the case, then then the number of available master nodes is not allowed to drop below the required quorum size. Fail the reconcile request with error if that is about to happen.
```

### DIFF
--- a/pkg/controller/nodemaintenance/check_quorum_size_test.go
+++ b/pkg/controller/nodemaintenance/check_quorum_size_test.go
@@ -1,0 +1,180 @@
+package nodemaintenance
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	policy "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	k8sfakeclient "k8s.io/client-go/kubernetes/fake"
+)
+
+func areErrorsEqual(err1, err2 error) bool {
+	return (err1 != nil && err2 != nil && err1.Error() == err2.Error()) ||
+		(err1 == nil && err2 == nil)
+}
+
+var _ = Describe("checkQuorumValidity", func() {
+	type TestCaseDefinition struct {
+		info            string
+		currentHealthy  int32
+		requiredHealthy int32
+		disruptedPods   map[string]metav1.Time
+		isValidQuorum   bool
+	}
+	It("check fallback option of quorum check", func() {
+		testCases := []TestCaseDefinition{
+			{
+				info:            "quorum broken",
+				currentHealthy:  2,
+				requiredHealthy: 3,
+				isValidQuorum:   false,
+			},
+			{
+
+				info:            "quorum valid",
+				currentHealthy:  3,
+				requiredHealthy: 2,
+				isValidQuorum:   true,
+			},
+			{
+				info:            "quorum not broken with dirsupted pods",
+				currentHealthy:  4,
+				requiredHealthy: 2,
+				disruptedPods: map[string]metav1.Time{
+					"disruptedPod": metav1.Time{},
+				},
+				isValidQuorum: true,
+			},
+			{
+				info:            "quorum broken with dirsupted pods",
+				currentHealthy:  4,
+				requiredHealthy: 4,
+				disruptedPods: map[string]metav1.Time{
+					"disruptedPod": metav1.Time{},
+				},
+				isValidQuorum: false,
+			},
+		}
+		for _, c := range testCases {
+			By("test:" + c.info)
+			Context(c.info, func() {
+
+				masterNode := &corev1.Node{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "masternode",
+						Labels: map[string]string{
+							MasterNodeLabel: "",
+						},
+					},
+					Spec: corev1.NodeSpec{},
+				}
+				objs := []runtime.Object{
+					&policy.PodDisruptionBudgetList{
+						Items: []policy.PodDisruptionBudget{
+							policy.PodDisruptionBudget{
+								ObjectMeta: metav1.ObjectMeta{
+									Namespace: OpenshiftMachineConfigNamespace,
+								},
+								Spec: policy.PodDisruptionBudgetSpec{
+									MinAvailable: &intstr.IntOrString{
+										Type:   intstr.Int,
+										IntVal: int32(c.requiredHealthy),
+									},
+								},
+								Status: policy.PodDisruptionBudgetStatus{
+									CurrentHealthy: c.currentHealthy,
+									DisruptedPods:  c.disruptedPods,
+								},
+							},
+						},
+					},
+					masterNode,
+				}
+				cs := k8sfakeclient.NewSimpleClientset(objs...)
+				isValidQuorum, isFallbackCheck, _ := checkValidQuorum(cs, masterNode)
+				Expect(isValidQuorum).To(Equal(c.isValidQuorum))
+				Expect(isFallbackCheck).To(Equal(false))
+
+			})
+		}
+	})
+})
+
+var _ = Describe("checkQuorumValidityFallback", func() {
+	type TestCaseDefinition struct {
+		info            string
+		currentHealthy  int32
+		requiredHealthy int32
+		disruptedPods   map[string]metav1.Time
+		isValidQuorum   bool
+	}
+	It("check fallback option of quorum check", func() {
+		testCases := []TestCaseDefinition{
+			{
+				info:            "quorum broken",
+				currentHealthy:  2,
+				requiredHealthy: 3,
+				isValidQuorum:   false,
+			},
+			{
+
+				info:            "quorum valid",
+				currentHealthy:  3,
+				requiredHealthy: 2,
+				isValidQuorum:   true,
+			},
+			{
+				info:            "quorum not broken with dirsupted pods",
+				currentHealthy:  4,
+				requiredHealthy: 2,
+				disruptedPods: map[string]metav1.Time{
+					"disruptedPod": metav1.Time{},
+				},
+				isValidQuorum: true,
+			},
+		}
+		for _, c := range testCases {
+			By("test:" + c.info)
+			Context(c.info, func() {
+
+				masterNode := &corev1.Node{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "masternode",
+						Labels: map[string]string{
+							MasterNodeLabel: "",
+						},
+					},
+					Spec: corev1.NodeSpec{},
+				}
+				objs := []runtime.Object{
+							&policy.PodDisruptionBudget{
+								ObjectMeta: metav1.ObjectMeta{
+									Name: "etcpdb",
+									Namespace: NodeMaintenanceOperatorNamespace,
+								},
+								Spec: policy.PodDisruptionBudgetSpec{
+									MinAvailable: &intstr.IntOrString{
+										Type:   intstr.Int,
+										IntVal: int32(c.requiredHealthy),
+									},
+								},
+								Status: policy.PodDisruptionBudgetStatus{
+									CurrentHealthy: c.currentHealthy,
+									DisruptedPods:  c.disruptedPods,
+								},
+							},
+							masterNode,
+				}
+				cs := k8sfakeclient.NewSimpleClientset(objs...)
+				isValidQuorum, isFallbackCheck, _ := checkValidQuorum(cs, masterNode)
+				Expect(isValidQuorum).To(Equal(c.isValidQuorum))
+				Expect(isFallbackCheck).To(Equal(true))
+
+			})
+		}
+	})
+})
--- a/pkg/controller/nodemaintenance/check_quorum_size_test.go
+++ b/pkg/controller/nodemaintenance/check_quorum_size_test.go
@@ -103,4 +103,3 @@ var _ = Describe("checkQuorumValidity", func() {
 	})
 })
 
-
--- a/pkg/controller/nodemaintenance/check_quorum_size_test.go
+++ b/pkg/controller/nodemaintenance/check_quorum_size_test.go
@@ -95,86 +95,12 @@ var _ = Describe("checkQuorumValidity", func() {
 					masterNode,
 				}
 				cs := k8sfakeclient.NewSimpleClientset(objs...)
-				isValidQuorum, isFallbackCheck, _ := checkValidQuorum(cs, masterNode)
+				isValidQuorum, err := checkValidQuorum(cs, masterNode)
 				Expect(isValidQuorum).To(Equal(c.isValidQuorum))
-				Expect(isFallbackCheck).To(Equal(false))
-
+				Expect(err).NotTo(HaveOccurred())
 			})
 		}
 	})
 })
 
-var _ = Describe("checkQuorumValidityFallback", func() {
-	type TestCaseDefinition struct {
-		info            string
-		currentHealthy  int32
-		requiredHealthy int32
-		disruptedPods   map[string]metav1.Time
-		isValidQuorum   bool
-	}
-	It("check fallback option of quorum check", func() {
-		testCases := []TestCaseDefinition{
-			{
-				info:            "quorum broken",
-				currentHealthy:  2,
-				requiredHealthy: 3,
-				isValidQuorum:   false,
-			},
-			{
 
-				info:            "quorum valid",
-				currentHealthy:  3,
-				requiredHealthy: 2,
-				isValidQuorum:   true,
-			},
-			{
-				info:            "quorum not broken with dirsupted pods",
-				currentHealthy:  4,
-				requiredHealthy: 2,
-				disruptedPods: map[string]metav1.Time{
-					"disruptedPod": metav1.Time{},
-				},
-				isValidQuorum: true,
-			},
-		}
-		for _, c := range testCases {
-			By("test:" + c.info)
-			Context(c.info, func() {
-
-				masterNode := &corev1.Node{
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "masternode",
-						Labels: map[string]string{
-							MasterNodeLabel: "",
-						},
-					},
-					Spec: corev1.NodeSpec{},
-				}
-				objs := []runtime.Object{
-							&policy.PodDisruptionBudget{
-								ObjectMeta: metav1.ObjectMeta{
-									Name: "etcpdb",
-									Namespace: NodeMaintenanceOperatorNamespace,
-								},
-								Spec: policy.PodDisruptionBudgetSpec{
-									MinAvailable: &intstr.IntOrString{
-										Type:   intstr.Int,
-										IntVal: int32(c.requiredHealthy),
-									},
-								},
-								Status: policy.PodDisruptionBudgetStatus{
-									CurrentHealthy: c.currentHealthy,
-									DisruptedPods:  c.disruptedPods,
-								},
-							},
-							masterNode,
-				}
-				cs := k8sfakeclient.NewSimpleClientset(objs...)
-				isValidQuorum, isFallbackCheck, _ := checkValidQuorum(cs, masterNode)
-				Expect(isValidQuorum).To(Equal(c.isValidQuorum))
-				Expect(isFallbackCheck).To(Equal(true))
-
-			})
-		}
-	})
-})
--- a/pkg/controller/nodemaintenance/master_quorum_check.go
+++ b/pkg/controller/nodemaintenance/master_quorum_check.go
@@ -14,29 +14,29 @@ import (
 
 const OpenshiftMachineConfigNamespace = "openshift-machine-config-operator"
 
-func checkValidQuorum(client kubernetes.Interface, node *corev1.Node) (bool, bool, error) {
+func checkValidQuorum(client kubernetes.Interface, node *corev1.Node) (bool, error) {
 	if isMasterNode(client, node) {
-		validQuorum, fallbackPDB, err := checkValidQuorumPdb(client)
+		validQuorum, err := checkValidQuorumPdb(client)
 		if err != nil {
-			return false, false, err
+			return false, err
 		}
 		log.Infof("node %s is a master. quorum status: %t", node.Name, validQuorum)
-		return validQuorum, fallbackPDB, nil
+		return validQuorum, nil
 	} else {
 		log.Infof("node %s is not a master", node.Name)
 	}
-	return true, false, nil
+	return true, nil
 }
 
-func findMcoPDBForEtcd(client kubernetes.Interface) (*policy.PodDisruptionBudget,bool,error) {
+func findMcoPDBForEtcd(client kubernetes.Interface) (*policy.PodDisruptionBudget,error) {
 	pdbList, err := client.Policy().PodDisruptionBudgets(OpenshiftMachineConfigNamespace).List(metav1.ListOptions{})
 	if err != nil {
-		return nil, false, fmt.Errorf("can't get mco pdb: %v", err)
+		return nil, fmt.Errorf("can't get mco pdb: %v", err)
 	}
 	if pdbList.Size() == 0  || pdbList.Items == nil || len(pdbList.Items) == 0 {
-		return nil, true, fmt.Errorf("mco pdb is empty")
+		return nil, fmt.Errorf("mco pdb is empty")
 	}
-	return &pdbList.Items[0], false, nil
+	return &pdbList.Items[0], nil
 }
 
 func isMasterNode(client kubernetes.Interface, node *corev1.Node) bool {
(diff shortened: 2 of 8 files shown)


pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -204,6 +222,48 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+	masterNodes := 0
+	availableMasterNodes := 0
+
+	n, err := r.drainer.Client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: MasterNodeLabelSelector})
+	if err != nil {
+		return -1, -1, err
+	}
+	for _, node := range n.Items {
+		if !node.Spec.Unschedulable {
+			availableMasterNodes += 1
```
n1r1, line 254: An admin might mark a node as unschedulable. This doesn't mean it's not available for quorum

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -204,6 +222,48 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileNodeMaintenance) checkQuorumValidity(node *corev1.Node) (bool, error) {
```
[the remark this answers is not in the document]
n1r1: actually, it would be nice if you can add docs to all the newly added non-trivial functions

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -177,6 +182,19 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+
+	validQuorum, err = r.checkQuorumValidity(node)
+	if err != nil {
+		reqLogger.Errorf("Can't check quorum validity: %v", err)
+		return r.reconcileAndError(instance, err)
+	}
+	if !validQuorum {
+		err = fmt.Errorf(QuorumViolationErrorMsg)
+		reqLogger.Error(err)
+		return r.reconcileAndError(instance, err)
```
n1r1, line 195: I'm not sure that we want to return a reconcile err in this case, it will lead to hot loop, since error means requeue and I don't expect that this will succeed immediately after.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -204,6 +222,48 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+			availableMasterNodes += 1
+		}
+		masterNodes += 1
+	}
+
+	return masterNodes, availableMasterNodes, nil
+}
+
+func isMasterNode(node *corev1.Node) bool {
+	_, isMaster := node.ObjectMeta.Labels[MasterNodeLabel]
```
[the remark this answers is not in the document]
MoserMichael: ObjectMeta is not a pointer field; I can't check it for nil. It is an object reference that can't be nil. Also the  Node structure in ./vendor/k8s.io/api/core/v1/types.go doesn't have a Labels field.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
MoserMichael: added comment to the method.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -204,6 +222,48 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+	}
+
+	masterNodes, availableMasterNodes, err := r.countMasterNodes()
+	if err != nil {
+		return false, err
+	}
+
+	quorumValid := availableMasterNodes-1 >= (masterNodes/2 + 1)
+
+	log.Infof("Node %s is a master, but quorum is still maintained. (availableMasters %d totalMasters %d", node.Name, availableMasterNodes, masterNodes)
```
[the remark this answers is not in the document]
MoserMichael: changed the log message.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: That is generally true, but for this exercise I cant tell if the node has been marked as unschedulable for the purpose of takeing it into maintenance (i.e. to shut it down) or not.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: fixed, changed to reschedule after five minutes.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -204,6 +222,48 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+
+	return quorumValid, nil
+}
+
+func (r *ReconcileNodeMaintenance) countMasterNodes() (int, int, error) {
+	masterNodes := 0
+	availableMasterNodes := 0
+
+	n, err := r.drainer.Client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: MasterNodeLabelSelector})
```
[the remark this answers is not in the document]
MoserMichael: That is interesting. I would like to learn more about that. The bugzilla report for this PR (link at the top) advised me to look at https://github.com/openshift/etcd-quorum-guard - the e2e test in this repository were all counting the master nodes by listing all nodes with that label selector.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
n1r1: > ObjectMeta is not a pointer field; I can't check it for nil. It is an object reference that can't be nil

You can check if `node.ObjectMeta.Labels == nil`

> Also the Node structure in ./vendor/k8s.io/api/core/v1/types.go doesn't have a Labels field.

There's a feature in golang named field promotion which allows you to access the `Labels` without specifying `ObjectMeta`.
Anyway, this is just a syntactic sugar. You don't have to change that.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
n1r1: It is still incorrect. Now it's the other way around.
The log says that quorum will be violated but this is not always the case (i.e. if `quorumValid == true`)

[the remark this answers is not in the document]
MoserMichael: fixed again.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
MoserMichael: fixed again.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
n1r1: What about adding an annotation to the node when it's being unschedulable by NMO?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
n1r1: When baremetal Machine turns unhealthy, we shut it down (using hardware management interface), remove the node from the cluster, and power it back on.
It makes me think that we need to be more careful with masters remediation. It feels like MCO/NMO might drain a master node, while machine-remediation might take other master down. 
/cc @beekhof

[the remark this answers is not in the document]
MoserMichael: How do you remove the node from the cluster? I understand from your previous comment that it is still returned by the go-client List call, How can I know if a node has been 'removed from the cluster' ?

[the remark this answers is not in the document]
n1r1: See implementation [here](https://github.com/openshift/cluster-api-provider-baremetal/blob/master/pkg/cloud/baremetal/actuators/machine/actuator.go#L952).

I don't expect it to be returned by the client.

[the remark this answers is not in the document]
MoserMichael: ok, so once you delete the node object via client-go and it gets deleted from etcd, then it will not be picked up by the List selector, so I don't see a problem to begin with.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: I don't know if you want to do that: you get race conditions if a node is being set for maintenance and hasn't reached that stage yet.

MoserMichael: After discussion with @beekhof  - he says we should rather err on the side of caution, so leave it as it is now.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
MoserMichael: Perhaps we could add a placeholder object upon deleting a master node, and delete the placeholder object when the deleted master node comes up again, then I could count the placeholder objects and add it to the number of active master nodes. We could also avoid conflicts by storing the node name in the placeholder object, so that nodes will not be counted twice.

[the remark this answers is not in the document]
n1r1: I think the optimal approach would be getting this from etcd-operator or any other entity in the system that has the knowledge. Not sure if this is possible though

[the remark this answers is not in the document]
MoserMichael: They gather this knowledge from the raft consensus algorithm. I am not sure that a simple operator wants to be exposed to the internals of the etcd raft implementation; these things might change suddenly between versions of opernshift/kubernetes and we will be the last to know they changed.

hack/coverage.sh
```diff
@@ -13,9 +13,28 @@ declare -a EXCLUDE_FILES_FROM_COVERAGE=("nodemaintenance_controller_init.go")
 # delete coverage files (if present)
 find . -name ${COVERAGE_FILE} | xargs rm -f
 
-# run ginkgo with coverage result line
-${GINKGO} ${GINKGO_ARGS} ./pkg/ ./cmd/ | sed  '/coverage:.*$/d'
-GSTAT=${PIPESTATUS[0]}
+while [ true ]; do
+
+		# run ginkgo with coverage result line
+		${GINKGO} ${GINKGO_ARGS} ./pkg/ ./cmd/ 2>&1 | sed  '/coverage:.*$/d'  | tee msg
```
beekhof, line 19: If ginkgo is failing for an unknown reason, maybe it makes sense to stop erasing parts of it's output incase it holds some clue to the failures.

In any case, can we remove this change and continue the conversation in the other PR you created to address this? 
   https://github.com/kubevirt/node-maintenance-operator/pull/78/files/c0c1a0bd5bcddd27543089dbf198adea940a76c3#r424158734

pkg/controller/nodemaintenance/master_quorum_check.go
```diff
@@ -0,0 +1,98 @@
...
+
+	masterNodes := 0
+	availableMasterNodes := 0
+
+	n, err := client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: MasterNodeLabelSelector})
+	if err != nil {
+		return -1, -1, err
+	}
+	for _, node := range n.Items {
+		if !node.Spec.Unschedulable {
```
beekhof, line 83: I believe Mike L noticed that not all failed nodes get marked as `Unschedulable`.
You might need to find something in node.Status that is a better indicator.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: Doesn't this result in the error being logged twice?

pkg/controller/nodemaintenance/master_quorum_check.go
MoserMichael: one possibility is to check if kubelet has updated the lease for that node recently, but that won't kick in for quite some time., until kubelet rececks the lease that is, and it is probably not very reliable when there is a large number of nodes.

MoserMichael: or we just might issue a warning and not check the condition if there is no PDB; so that there is no fallback option. Is that preferable?

hack/coverage.sh
MoserMichael: removed

pkg/controller/nodemaintenance/master_quorum_check.go
MoserMichael: Maybe it's enough in any case to check node.Status.Phase == v1.NodeTerminated

However right now I don't see where this field value is set in kubernetes.

./pkg/apis/core/types.go
	// NodeTerminated means the node has been removed from the cluster.
	NodeTerminated NodePhase = "Terminated"

beekhof: https://github.com/kubernetes/kubernetes/pull/30005/files suggests that's not a good idea:

> The field is never populated, and now is deprecated.

You could either try to infer how many are healthy by looking at any reason the node might not hosting Pods (ie. look at taints too), or re-implement PDB by counting the number of etcd pods.

Probably better to do neither and just create a PDB in our own namespace (with the same definition) if an official one doesn't exist.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: reconcileAndError doesn't log the error that is passed as the second argument; at least not in this branch;

MoserMichael: I thought that the event of detected possible quorum violation  had to be logged with the logger as it is a serious event. Please correct me if I am wrong.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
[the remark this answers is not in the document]
MoserMichael: now the PR is checking for the MCO PDB. if that doesn't exist then we still use this code as a fallback option.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: now we are looking at taints instead of the unschedulable property , which is more reliable.

pkg/controller/nodemaintenance/master_quorum_check.go
MoserMichael: pushed the taint based solution.

ovirt-infra:
All tests passed

pkg/controller/nodemaintenance/master_quorum_check.go
MoserMichael: pushed fallback solution based on pdb.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: now using pdb based solution instead.

pkg/controller/nodemaintenance/check_quorum_size_test.go
```diff
@@ -0,0 +1,171 @@
...
+			},
+			{
+
+				info:            "quorum valid",
+				currentHealthy:  3,
+				requiredHealthy: 2,
+				isValidQuorum:   true,
+			},
+			{
+				info:            "quorum not broken with dirsupted pods",
```
beekhof, line 44: A disrupted pod effectively just subtracts 1 from the currentHealthy count.
So I don't think this is testing the right condition.

I believe you want `requiredHealthy: 4` and look for `isValidQuorum: false`

pkg/controller/nodemaintenance/master_quorum_check.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func checkValidQuorumPdb(client kubernetes.Interface) (bool, bool, error) {
+
+	fallbackPDB := false
+	pdb, pdbNotFound, err := findMcoPDBForEtcd(client)
+	if pdbNotFound {
+		fallbackPDB = true
+		pdb, err = createFallbackPDB(client)
+	}
```
beekhof, line 90: Remember when I mentioned creating this as part of the test harness, not in Go code?

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -283,6 +304,13 @@ func (r *ReconcileNodeMaintenance) initMaintenanceStatus(nm *nodemaintenanceapi.
 	return nil
 }
 
+func (r *ReconcileNodeMaintenance) reconcileAndErrorAfter(nm *nodemaintenanceapi.NodeMaintenance, err error) (reconcile.Result, error) {
```
beekhof, line 307: Why does this function exist?  It does not appear to be used anywhere?

test/e2e/nodemaintenance_test.go
```diff
@@ -151,27 +173,157 @@ func nodeMaintenanceTest(t *testing.T, f *framework.Framework, ctx *framework.Te
...
+		if nm.Status.LastError != nmo.QuorumViolationErrorMsg {
+			t.Logf("%s: Running, with wrong error: %s", time.Now().Format("2006-01-02 15:04:05.000000"), nm.Status.LastError)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		showDeploymentStatus(t, f, fmt.Errorf("%s: Failed to verify running phase: %v", time.Now().Format("2006-01-02 15:04:05.000000"), err))
+	}
+
+	t.Logf("correct error has been observed")
```
beekhof, line 206: So the e2e tests never actually succeed in putting node into maintenance mode anymore?

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: removed

pkg/controller/nodemaintenance/master_quorum_check.go
MoserMichael: done

pkg/controller/nodemaintenance/check_quorum_size_test.go
MoserMichael: UpAfterNMO := requiredHealthy(4) - len(disruptedPods) - 1 /* the node to be taken offline */ == 2.

UpAfterNMO (2) must be >= requiredHealthy (2) , therefore the test must pass.

pkg/controller/nodemaintenance/master_quorum_check.go
MoserMichael: done, also moved the this e2e test case  into a separate file.

pkg/controller/nodemaintenance/check_quorum_size_test.go
beekhof: I believe we discussed previously, that the test will pass even if `disruptedPods` is `nil`.  
That makes it a bad test.

If you set `requiredHealthy: 4` and look for `isValidQuorum: false`, then you validate that `disruptedPods` are being correctly taken into account.

MoserMichael: added this test case "

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

go.mod
```diff
@@ -27,28 +26,25 @@ require (
...
 	github.com/pborman/uuid v0.0.0-20180906182336-adf5a7427709 // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
-	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90 // indirect
-	github.com/prometheus/common v0.2.0 // indirect
-	github.com/prometheus/procfs v0.0.0-20190225181712-6ed1f7e10411 // indirect
 	github.com/sergi/go-diff v1.1.0 // indirect
-	github.com/sirupsen/logrus v1.3.0
-	github.com/spf13/pflag v1.0.3
+	github.com/sirupsen/logrus v1.4.2
+	github.com/spf13/pflag v1.0.5
```
beekhof, line 37: Why are all these versions changed?
I can see that the commit message says "update go.mod" but it lacks the most important information: why

pkg/controller/nodemaintenance/master_quorum_check.go
beekhof: I'm confused. You say done, but the logic that creates a PDB is still an active part of the flow.

MoserMichael: I misunderstoood that  we needed a fallback option if the operator is running on an old cluster (k8s without mco) or old cluster (os 3.11). Apparently that is incorrect, so I will delete.

MoserMichael: so if the node is a master node and we can't get the pdb of the mco, then remediation failed. Is that correct?

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

go.mod
MoserMichael: removed the commit.

MoserMichael: removed the commit. I saw that go.mod file changed locally. This probably happened because I switched branches locally.

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

test/e2e/nodemaintenance_quorumsize_test.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, nil, err
+	}
+	return object, nsobject, nil
+}
+
+func checkQuorumSizeViolation(t *testing.T, f *framework.Framework, ctx *framework.TestCtx) error {
+
+	pdb, nsobject, err := createEtcPDB(t , f , ctx )
+	if err != nil {
+		showDeploymentStatus(t, f, fmt.Errorf("can't create test pdb :  %v", err))
```
beekhof, line 71: This isn't an error if the PDB already exists, which is what we expect on an OCP 4 cluster

test/e2e/nodemaintenance_quorumsize_test.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	err = deleteSimpleDeployment(t, f, ctx, namespace)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("test deployment deleted")
+	t.Logf("test checkQuorumSizeViolation completed successfully")
+
+	err = f.Client.Delete(goctx.TODO(), pdb)
```
beekhof, line 200: Assuming we created it.  Which again, should not be the case on OCP 4

MoserMichael: we create it in all the test environments (including ocp 4.3) If that weren't the case then  createEtcPDB would have failed at line #59 with error.IsAlreadyExist.

test/e2e/nodemaintenance_quorumsize_test.go
MoserMichael: we create the pdb it in all the test environments (including ocp 4.3) If that weren't the case then createEtcPDB would have failed at line #59 with error.IsAlreadyExist.

beekhof: CI isn't the only place these tests can/should/will run, and kubevirtci isn't the only CI that could be used.
Marc already has a todo item for getting access to the OCP CI

test/e2e/nodemaintenance_quorumsize_test.go
beekhof: CI isn't the only place these tests can/should/will run, and kubevirtci isn't the only CI that could be used.
Marc already has a todo item for getting access to the OCP CI

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

test/e2e/nodemaintenance_quorumsize_test.go
MoserMichael: Makes sense. Added 1f312fed09860b330d20443707b378df43b68a1c that deletes the objects if they were created for the purpose of this test.

test/e2e/nodemaintenance_quorumsize_test.go
MoserMichael: Makes sense. Added 1f312fe that deletes the objects if they were created for the purpose of this test.

test/e2e/nodemaintenance_quorumsize_test.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	err = deleteSimpleDeployment(t, f, ctx, namespace)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("test deployment deleted")
+	t.Logf("test checkQuorumSizeViolation completed successfully")
+
+	if pdbCreated {
```
beekhof, line 206:
```suggestion
	if pdb != nil {
```

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

test/e2e/nodemaintenance_quorumsize_test.go
MoserMichael: done

slintes:
Closing for #107 

/close

PR #77, opened 2020-05-08
Add support for admission webhook to validate CRD object create request.

MoserMichael:

This PR is supposed to solve issue https://bugzilla.redhat.com/show_bug.cgi?id=1825997

Need to validate the node name field prior to creating the CRD object. The objective is to prevent creation of CRD object with invalid node name. 
A validating admission webhook was added for this purpose

**Special notes for your reviewer:**

The PR adds an valiating admission webhook that validates CREATE requests, The purpose is to reject attempts to create a CRD instance with invalid node name.

The changes are in the following files.
,/Makefile
./cmd/manager/main.go
./deploy/role.yaml
./hack/coverage.sh
./hack/functest.sh
./pkg/controller/nodemaintenance/nodemaintenance_controller_init.go
./pkg/controller/nodemaintenanc/nodemaintenance_webhookaction.go
./pkg/webhooks/webhooks.go
./test/e2e/nodemaintenance_test.go

Please note that there is a general package for adding the webhook functionality ./pkg/webhooks/webhooks.go  whereas the actual nmo logic is in ./pkg/controller/nodemaintenanc/nodemaintenance_webhookaction.go
Setup of the webhook during operator startup is in ./pkg/controller/nodemaintenance/nodemaintenance_controller_init.go and ./cmd/manager/main.go

operator-sdk currently has no support for webhooks, so I had to run my own implementation.

Please note that there are no unit tests for this functionality, all testing is done in the e2e test;  To explain the reason first look at how webhooks are invoked:
 
- the api server reads the validatingwebhookconfiguration object that has a service reference as well as the CA bundle needed to communicate with the service that hosts the webhook; 
- there is a service object that defines how to route the IP traffic to the webhook service,  
- in the operator code there is a https server that hosts the webhook;

All in all these interactions are quite difficult to test with the fake client that are regularly used with unit testing, as they all involve quite a lot of tcp/ip conversations.

**Release note:**
```release-note
Need to validate the node name field prior to creating the CRD object. 
The objective is to prevent creation of CRD object with invalid node name. 
A validating admission webhook was added for this purpose
```


n1r1:
> The PR adds an valiating admission webhook that validates CREATE requests

Any reason not to validate PUT/PATCH requests as well? 
Does that even make sense to edit node name after CR creation? if not, we should block this as well I guess.

WDYT?

MoserMichael:
> > The PR adds an valiating admission webhook that validates CREATE requests
> 
> Any reason not to validate PUT/PATCH requests as well?
> Does that even make sense to edit node name after CR creation? if not, we should block this as well I guess.
> 
> WDYT?

the bugzilla issue that this PR is supposed to solve asked for verification of node name on creating a CRD object. If the bug report is changed then we will verify these as well.

MoserMichael:
was decided to use olm for webhooks. They create a certificate and mount it into the operator pod. Here in this PR we create our own self-signed certificate for that purpose. (i asked around and OLM also creates a self signed cert for this purpose, but that would be not our problem in case the system is reviewed for that purpose).

PR #78, opened 2020-05-10
fix intermittent failure to start unit test runner

MoserMichael:
Occasionally ginkgo test run fails to start, the error message in this
case is 'unable to combine coverage'
This condition is fixed by repeating the test run.

The full error log for a failed test run:

      quorum broken
      /home/mmoser/go/src/kubevirt.io/node-maintenance-operator/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:43
    [It] quorum broken
      /home/mmoser/go/src/kubevirt.io/node-maintenance-operator/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:43
    path is /home/mmoser/go/src/kubevirt.io/node-maintenance-operator
    Unable to read coverage file pkg/controller/nodemaintenance/cover.out to combine, open pkg/controller/nodemaintenance/cover.out: no such file or directory

Signed-off-by: Michael Moser <mmoser@redhat.com>

### DIFF
--- a/hack/coverage.sh
+++ b/hack/coverage.sh
@@ -4,7 +4,9 @@ GINKGO="$1"
 TARGETCOVERAGE="$2"
 
 COVERAGE_FILE=cover.out
-GINKGO_COVERAGE_ARGS="-cover -coverprofile=${COVERAGE_FILE} -outputdir=. --skipPackage ./vendor"
+
+#GINKGO_COVERAGE_ARGS="-coverpkg  ./,./pkg/controller/nodemaintenance -coverprofile=${COVERAGE_FILE} -outputdir=.  --skipPackage ./vendor"
+GINKGO_COVERAGE_ARGS="-cover -coverprofile=${COVERAGE_FILE} -outputdir=.  --skipPackage ./vendor"
 GINKGO_ARGS="-v -r --progress ${GINKGO_EXTRA_ARGS} ${GINKGO_COVERAGE_ARGS}"
 
 # source files excluded from ginkgo coverage report. these files are not used during the unit test and include code that is only relevant to the installed product.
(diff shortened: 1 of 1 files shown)


hack/coverage.sh
```diff
@@ -13,9 +14,31 @@ declare -a EXCLUDE_FILES_FROM_COVERAGE=("nodemaintenance_controller_init.go")
...
 
-# run ginkgo with coverage result line
-${GINKGO} ${GINKGO_ARGS} ./pkg/ ./cmd/ | sed  '/coverage:.*$/d'
-GSTAT=${PIPESTATUS[0]}
+REPEAT_COUNT=0
+MAX_REPEAT_COUNT=10
+while [[ $REPEAT_COUNT -lt $MAX_REPEAT_COUNT ]]; do
+
+		# run ginkgo with coverage result line
+		${GINKGO} ${GINKGO_ARGS} ./pkg/ ./cmd/ 2>&1 | sed  '/coverage:.*$/d'  | tee msg
```
n1r1, line 22: Instead of using `tee` you might want just to use `grep`, this will be helpful in order to avoid creating new files on disks that later could appear as changes in git

MoserMichael: fixed.

hack/coverage.sh
```diff
@@ -4,7 +4,8 @@ GINKGO="$1"
 TARGETCOVERAGE="$2"
 
 COVERAGE_FILE=cover.out
-GINKGO_COVERAGE_ARGS="-cover -coverprofile=${COVERAGE_FILE} -outputdir=. --skipPackage ./vendor"
+
+GINKGO_COVERAGE_ARGS="-cover -coverprofile=${COVERAGE_FILE} -outputdir=.  --skipPackage ./vendor"
```
beekhof, line 8: Can we drop the unnecessary whitespace changes please?

hack/coverage.sh
```diff
@@ -13,9 +14,32 @@ declare -a EXCLUDE_FILES_FROM_COVERAGE=("nodemaintenance_controller_init.go")
...
 
-# run ginkgo with coverage result line
-${GINKGO} ${GINKGO_ARGS} ./pkg/ ./cmd/ | sed  '/coverage:.*$/d'
-GSTAT=${PIPESTATUS[0]}
+REPEAT_COUNT=0
+MAX_REPEAT_COUNT=10
+while [[ $REPEAT_COUNT -lt $MAX_REPEAT_COUNT ]]; do
+
+		# run ginkgo with coverage result line
+		GINKGO_OUT=$(${GINKGO} ${GINKGO_ARGS} ./pkg/ ./cmd/ 2>&1 | sed  '/coverage:.*$/d')
```
beekhof, line 22: Are you sure that the `coverage:` lines do not contain any information about why this test is flaky?

MoserMichael: The lines taken out are from nodemaintenace_controller_init.go - these are functions that have zero coverage during unit testing. (like addition of controller to the controller runtime manager etc).
I checked that the problem happened without these changes at all, so i don't think there is a connection.

hack/coverage.sh
MoserMichael: fixed

hack/coverage.sh
MoserMichael: i get it without this trick too; see https://github.com/onsi/ginkgo/issues/417

MoserMichael: I actually haven't seen these failures of the test runner for quite some time now.
It might have been some  problem on the side of the build server environment.

beekhof:
Problem seems to have gone away

PR #91, opened 2020-06-21
whitespace fixes as precommit hook

MoserMichael:
- make check installs a pre-commit hook that reports whitespace issues (trailing whitespace or multiple empty lines). the pre-commit hook checks files that are about to be commited for whitespace problems.
- make whitespace-check now reports files with multiple empty lines as errors.
- make whitepace-commit : fixes whitespace issues in files added for commit with git add.
- while we are at here: also add git hook that sets the signoff message into the commit message, so that you don't have to do git commit --amend --signoff 

The rational for these changes is that it is more appropriate to fix trailing whitespaces and multiple empty lines when you are about to commit something

### DIFF
--- a/hack/precommit-hook.sh
+++ b/hack/precommit-hook.sh
@@ -2,14 +2,14 @@
 
 TOP_LEVEL=$(git rev-parse --show-toplevel)
 
-if [[ $1 == "setup" ]]; then
-	ln -fs  $TOP_LEVEL/hack/precommit-hook.sh $TOP_LEVEL/.git/hooks/pre-commit
-	exit 0
-fi
-
 if [[ $TOP_LEVEL == "" ]]; then
   echo "Error: this script must be called from a git repository"
   exit 1
+fi
+
+if [[ $1 == "setup" ]]; then
+	ln -fs  $TOP_LEVEL/hack/precommit-hook.sh $TOP_LEVEL/.git/hooks/pre-commit
+	exit 0
 fi
 
 # Redirect output to stderr.
--- a/hack/precommit-hook.sh
+++ b/hack/precommit-hook.sh
@@ -8,7 +8,9 @@ if [[ $TOP_LEVEL == "" ]]; then
 fi
 
 if [[ $1 == "setup" ]]; then
-	ln -fs  $TOP_LEVEL/hack/precommit-hook.sh $TOP_LEVEL/.git/hooks/pre-commit
+	if [[ ! -f $TOP_LEVEL/.git/hooks/pre-commit ]] && [[ ! -h $TOP_LEVEL/.git/hooks/pre-commit ]]; then
+		ln -s  $TOP_LEVEL/hack/precommit-hook.sh $TOP_LEVEL/.git/hooks/pre-commit
+	fi
 	exit 0
 fi
 
--- a/hack/precommit-hook.sh
+++ b/hack/precommit-hook.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+TOP_LEVEL=$(git rev-parse --show-toplevel)
+
+if [[ $1 == "setup" ]]; then
+	ln -fs  $TOP_LEVEL/hack/precommit-hook.sh $TOP_LEVEL/.git/hooks/pre-commit
+	exit 0
+fi
+
+if [[ $TOP_LEVEL == "" ]]; then
+  echo "Error: this script must be called from a git repository"
+  exit 1
+fi
+
+# Redirect output to stderr.
+exec 1>&2
+
+cd $TOP_LEVEL
+${TOP_LEVEL}/hack/whitespace.sh --check-precommit
+if [[ $? != 0 ]]; then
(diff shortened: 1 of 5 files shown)


ovirt-infra:

beekhof:
Happy to consider the extra whitespace checks, but not sure I'm keen on forcing the git hook to be everywhere.

@slintes thoughts?

hack/precommit-hook.sh
```diff
@@ -0,0 +1,23 @@
...
+
+TOP_LEVEL=$(git rev-parse --show-toplevel)
+
+if [[ $TOP_LEVEL == "" ]]; then
+  echo "Error: this script must be called from a git repository"
+  exit 1
+fi
+
+if [[ $1 == "setup" ]]; then
+	ln -fs  $TOP_LEVEL/hack/precommit-hook.sh $TOP_LEVEL/.git/hooks/pre-commit
```
slintes, line 11: I would prefer to not overwrite existing hooks by removing the `-f`. WDYT?

hack/prepare-commit-msg-hook.sh
```diff
@@ -0,0 +1,23 @@
...
+
+TOP_LEVEL=$(git rev-parse --show-toplevel)
+
+if [[ $TOP_LEVEL == "" ]]; then
+  echo "Error: this script must be called from a git repository"
+  exit 1
+fi
+
+if [[ $1 == "setup" ]]; then
+	ln -fs  $TOP_LEVEL/hack/prepare-commit-msg-hook.sh $TOP_LEVEL/.git/hooks/prepare-commit-msg
```
slintes, line 11: same here about overwriting

hack/prepare-commit-msg-hook.sh
```diff
@@ -0,0 +1,23 @@
...
+	exit 0
+fi
+
+UNAME=$(git config user.name)
+UMAIL=$(git config user.email)
+
+if [[ $UNAME != "" ]] && [[ $UMAIL != "" ]]; then
+	COMMIT_MSG_FILE="$1"
+	SIGNOFF_MSG="Signed-off-by: ${UNAME} <${UMAIL}>"
+	sed -ie "1i ${SIGNOFF_MSG}" "${COMMIT_MSG_FILE}"
```
slintes, line 21: I see 2 issues here:
- the signoff is added on top of the commit msg, it should be at the end
- people might have configured their IDE to always signoff commits automatically. This would add a 2nd signoff. I would prefer to just check if the commit is signed off and error if not

hack/whitespace.sh
```diff
@@ -9,14 +9,41 @@ function fix_one_file() {
...
 }
 
+function check_files {
+	while read fname; do
+		IDEAL_FILE=$(sed --follow-symlinks 's/[[:space:]]*$//' ${fname} | sed 'N;/^\n$/D;P;D;')
+		REAL_FILE=$(cat ${fname})
+
+		if [[ "$IDEAL_FILE" != "$REAL_FILE" ]]; then
+			echo "need whitespace fix for file $fname"
+			#diff -u $(echo "$REAL_FILE") $(echo "$IDEAL_FILE")
```
slintes, line 19: remove commented line please

hack/prepare-commit-msg-hook.sh
MoserMichael: done

hack/precommit-hook.sh
MoserMichael: i didn't find any pre-existing commit hooks on github, so that should b ok. done.

hack/prepare-commit-msg-hook.sh
MoserMichael: done

hack/whitespace.sh
MoserMichael: i put in the diff - now make whitespace-check is displaying the problem cass.

hack/prepare-commit-msg-hook.sh
```diff
@@ -0,0 +1,42 @@
...
+# check if commit message is empty
+HAS_SIGNOFF=$(grep -c 'Signed-off-by:' "${COMMIT_MSG_FILE}")
+
+if [[ $HAS_SIGNOFF == "0" ]]; then
+	UNAME=$(git config user.name)
+	UMAIL=$(git config user.email)
+
+	if [[ $UNAME != "" ]] && [[ $UMAIL != "" ]]; then
+
+		COMMIT_COMMENTS=$(sed -n 's/\(^#.*$\)/\1/p' "${COMMIT_MSG_FILE}")
```
slintes, line 28: I think you can skip this comment handling. I did some tests in my IDE and with git cli, the result is always the same.

However, the more I think about this, the more I think we should not automate the sign off, but only check for it and error if it is missing. Reason: at the end of the day signing off has legal impact, you are signing off that you agree to the DCO (https://developercertificate.org/). This shouldn't be automated but a deliberate action IMHO. @beekhof WDYT?

MoserMichael: makes sense, Will change it to validation only.

MoserMichael: done

PR #92, opened 2020-06-23
set owner ref on nmo CR to the node object

MoserMichael:
CR object must be deleted when node is being deleted. Therefore we need to add an owner ref from CR object to node.

The node ref is BlockOwnerDeletion = false, this requires us to do additional check when deleting the object. This creates the cleanup scenario:  referred to node has been deleted but lease hasn't. 
 (or should we rather add a owner ref from lease object to the node?) Didn't see anyone setting the owner ref of node lease in any usage of leases).

### DIFF
--- a/pkg/controller/nodemaintenance/nodemaintenance_controller.go
+++ b/pkg/controller/nodemaintenance/nodemaintenance_controller.go
@@ -256,8 +256,8 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	return reconcile.Result{}, nil
 }
 
-func makeBoolRef(val bool) (*bool) {
-	return &val;
+func makeBoolRef(val bool) *bool {
+	return &val
 }
 
 func setOwnerRefToNode(instance *nodemaintenanceapi.NodeMaintenance, node *corev1.Node) {
--- a/pkg/controller/nodemaintenance/nodemaintenance_controller.go
+++ b/pkg/controller/nodemaintenance/nodemaintenance_controller.go
@@ -256,35 +256,31 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	return reconcile.Result{}, nil
 }
 
+func makeBoolRef(val bool) (*bool) {
+	return &val;
+}
 
-func setOwnerRefToNode(instance *nodemaintenanceapi.NodeMaintenance, node *corev1.Node) (bool) {
+func setOwnerRefToNode(instance *nodemaintenanceapi.NodeMaintenance, node *corev1.Node) {
 
-	objectRefIsSet := false
 	for _, ref := range instance.ObjectMeta.GetOwnerReferences() {
 		if ref.APIVersion == node.TypeMeta.APIVersion && ref.Kind == node.TypeMeta.Kind && ref.Name == node.ObjectMeta.GetName() && ref.UID == node.ObjectMeta.GetUID() {
-			objectRefIsSet = true
-			break
+			return
 		}
 	}
 
-	if !objectRefIsSet {
-		log.Info("setting owner ref to node")
+	log.Info("setting owner ref to node")
 
-		blockOwnerDeletion := false
-		controllerDeletion := false
-		nodeMeta := node.TypeMeta
-		ref := metav1.OwnerReference{
-				APIVersion:         nodeMeta.APIVersion,
-				Kind:               nodeMeta.Kind,
-				Name:               node.ObjectMeta.GetName(),
-				UID:                node.ObjectMeta.GetUID(),
-				BlockOwnerDeletion: &blockOwnerDeletion, // false - deletion of node doesn't have to wait for nmo CR deletion
-				Controller:         &controllerDeletion, // false - can delete nmo without deleting the node
-			}
+	nodeMeta := node.TypeMeta
+	ref := metav1.OwnerReference{
+			APIVersion:         nodeMeta.APIVersion,
+			Kind:               nodeMeta.Kind,
+			Name:               node.ObjectMeta.GetName(),
+			UID:                node.ObjectMeta.GetUID(),
+			BlockOwnerDeletion: makeBoolRef(false),
+			Controller:         makeBoolRef(false),
+		}
 
-		instance.ObjectMeta.SetOwnerReferences( append(instance.ObjectMeta.GetOwnerReferences(), ref) )
-	}
-	return !objectRefIsSet
+	instance.ObjectMeta.SetOwnerReferences( append(instance.ObjectMeta.GetOwnerReferences(), ref) )
 }
 
 func (r *ReconcileNodeMaintenance) obtainLease(node *corev1.Node) (bool, error) {
--- a/pkg/controller/nodemaintenance/nodemaintenance_controller.go
+++ b/pkg/controller/nodemaintenance/nodemaintenance_controller.go
@@ -169,9 +169,9 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 		reqLogger.Infof("Deletion timestamp not zero")
 
 		// The object is being deleted
-		if ContainsString(instance.ObjectMeta.Finalizers, nodemaintenanceapi.NodeMaintenanceFinalizer) {
+		if ContainsString(instance.ObjectMeta.Finalizers, nodemaintenanceapi.NodeMaintenanceFinalizer) || ContainsString(instance.ObjectMeta.Finalizers, metav1.FinalizerOrphanDependents)  {
 			// Stop node maintenance - uncordon and remove live migration taint from the node.
-			if err := r.stopNodeMaintenance(instance.Spec.NodeName); err != nil {
+			if err := r.stopNodeMaintenanceOnDeletion(instance.Spec.NodeName); err != nil {
 				reqLogger.Infof("error stopping node maintenance: %v", err)
 				if errors.IsNotFound(err) == false {
 					return r.reconcileAndError(instance, err)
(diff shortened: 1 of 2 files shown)


ovirt-infra:

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -254,6 +256,37 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+		blockOwnerDeletion := false
+		controllerDeletion := false
+		nodeMeta := node.TypeMeta
+		ref := metav1.OwnerReference{
+				APIVersion:         nodeMeta.APIVersion,
+				Kind:               nodeMeta.Kind,
+				Name:               node.ObjectMeta.GetName(),
+				UID:                node.ObjectMeta.GetUID(),
+				BlockOwnerDeletion: &blockOwnerDeletion, // false - deletion of node doesn't have to wait for nmo CR deletion
+				Controller:         &controllerDeletion, // false - can delete nmo without deleting the node
```
beekhof, line 282: I think the comment here is wrong.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -297,9 +330,18 @@ func (r *ReconcileNodeMaintenance) stopNodeMaintenanceImp(node *corev1.Node ) er
 	return nil
 }
 
-func (r *ReconcileNodeMaintenance) stopNodeMaintenance(nodeName string) error {
+func (r *ReconcileNodeMaintenance) stopNodeMaintenanceOnDeletion(nodeName string) error {
 	node, err := r.fetchNode(nodeName)
 	if err != nil {
+		// if CR is gathered as result of garbage collection: the node may have been deleted, but the CR has not yet been deleted, still we must clean up the lease!
+		if errors.IsNotFound(err) {
+			if r.isLeaseSupported {
```
beekhof, line 338: I would normally expect that check to happen inside invalidateLease()

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -254,6 +256,37 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	return reconcile.Result{}, nil
 }
 
+
+func setOwnerRefToNode(instance *nodemaintenanceapi.NodeMaintenance, node *corev1.Node) (bool) {
+
+	objectRefIsSet := false
+	for _, ref := range instance.ObjectMeta.GetOwnerReferences() {
+		if ref.APIVersion == node.TypeMeta.APIVersion && ref.Kind == node.TypeMeta.Kind && ref.Name == node.ObjectMeta.GetName() && ref.UID == node.ObjectMeta.GetUID() {
+			objectRefIsSet = true
```
slintes, line 265: You can just `return false` here, and remove the `objectRefIsSet` var and the following `if` condition

pkg/controller/nodemaintenance/nodemaintenance_controller.go
slintes: Yes, also the name `controllerDeletion` does not make sense.
From the doc of this field: `If true, this reference points to the managing controller.`

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -254,6 +256,37 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
...
+				Kind:               nodeMeta.Kind,
+				Name:               node.ObjectMeta.GetName(),
+				UID:                node.ObjectMeta.GetUID(),
+				BlockOwnerDeletion: &blockOwnerDeletion, // false - deletion of node doesn't have to wait for nmo CR deletion
+				Controller:         &controllerDeletion, // false - can delete nmo without deleting the node
+			}
+
+		instance.ObjectMeta.SetOwnerReferences( append(instance.ObjectMeta.GetOwnerReferences(), ref) )
+	}
+	return !objectRefIsSet
```
slintes, line 287: The return value isn't evaluated in the actual code, so it can be removed.
You should not add a return value just for the test IMHO.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
slintes: agree, but probably out of scope of this PR

pkg/controller/nodemaintenance/nodemaintenance_controller_test.go
```diff
@@ -106,6 +106,25 @@ var _ = Describe("updateCondition", func() {
...
 			Expect(maintanance.Status.TotalPods).To(Equal(2))
 		})
+		It("owner ref should be set properly", func() {
+			r.initMaintenanceStatus(nm)
+			maintanance := &nodemaintenanceapi.NodeMaintenance{}
+			err := cl.Get(context.TODO(), req.NamespacedName, maintanance)
+			node, err := cs.CoreV1().Nodes().Get("node01", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			hasSetRef := setOwnerRefToNode(maintanance, node)
+			Expect(hasSetRef).To(Equal(true))
```
slintes, line 116: As mentioned before: the actual code does not use the return value so I think it should be removed.
In this test it is enough to check that the owner ref is set correctly, as it is done in the following lines.

general hint: `To(Equal(true))` can be `To(BeTrue())`

MoserMichael: return value of stOwnerRefToNode has been removed.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: comments removed, temporary variables removed.

pkg/controller/nodemaintenance/nodemaintenance_controller_test.go
MoserMichael: done, return value removed.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
beekhof: I'd rather not introduce it wrong in the first place :)

slintes: Yes, but is was already introduced, it's not a problem of this RP IMHO.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
```diff
@@ -254,6 +256,33 @@ func (r *ReconcileNodeMaintenance) Reconcile(request reconcile.Request) (reconci
 	return reconcile.Result{}, nil
 }
 
+func makeBoolRef(val bool) (*bool) {
+	return &val;
```
slintes, line 260: no need for () around the return value above, and the semicolon here

pkg/controller/nodemaintenance/nodemaintenance_controller_test.go
```diff
@@ -106,6 +106,23 @@ var _ = Describe("updateCondition", func() {
...
 			Expect(maintanance.Status.TotalPods).To(Equal(2))
 		})
+		It("owner ref should be set properly", func() {
+			r.initMaintenanceStatus(nm)
+			maintanance := &nodemaintenanceapi.NodeMaintenance{}
+			err := cl.Get(context.TODO(), req.NamespacedName, maintanance)
+			node, err := cs.CoreV1().Nodes().Get("node01", metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			setOwnerRefToNode(maintanance, node)
+			Expect(err).NotTo(HaveOccurred())
```
slintes, line 116: there is no new err to check

pkg/controller/nodemaintenance/nodemaintenance_controller_test.go
```diff
@@ -106,6 +106,23 @@ var _ = Describe("updateCondition", func() {
 			Expect(maintanance.Status.EvictionPods).To(Equal(2))
 			Expect(maintanance.Status.TotalPods).To(Equal(2))
 		})
+		It("owner ref should be set properly", func() {
+			r.initMaintenanceStatus(nm)
+			maintanance := &nodemaintenanceapi.NodeMaintenance{}
+			err := cl.Get(context.TODO(), req.NamespacedName, maintanance)
+			node, err := cs.CoreV1().Nodes().Get("node01", metav1.GetOptions{})
```
slintes, line 113: I see it is done at other places as well, but you should not hardcode node names.
The test can and will run on other CIs than KubeVirtCI with differenent node names.
I leave it up to you if you change it in this PR (at all places please) or in a follow up.

MoserMichael: this is a unit test, it doesn't run on any CI, in any event this is not related to any real nodes.

pkg/controller/nodemaintenance/nodemaintenance_controller.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller_test.go
MoserMichael: done

pkg/controller/nodemaintenance/nodemaintenance_controller_test.go
slintes: oh yes, of course, sorry :slightly_smiling_face:

PR #96, opened 2020-07-01
Coordinate between different NMO instances

MoserMichael:
Need to ensure one active copy of node maint. operator running in different namespaces, this PR adds a variant of leader election that creates a lock object in the default namespace, so that the lock is visible to all instances.

Currently all versions of the operator are treated the same - no mechanisms has been implemented yet that would  allow to handle the more advanced version of the operator with precedence over less advanced versions.

e2e test of the feature were added.

Note that if this solution is accepted then the same leader check must be backported to all relevant versions.

**Release note:**
```release-note 
two installations of NMO in different namespaces may not reconcile the same CR object.
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -10,8 +10,8 @@ import (
 	"kubevirt.io/node-maintenance-operator/pkg/apis"
 	"kubevirt.io/node-maintenance-operator/pkg/controller"
 	"kubevirt.io/node-maintenance-operator/pkg/controller/nodemaintenance"
+	"kubevirt.io/node-maintenance-operator/pkg/controller/nodemaintenance/leader"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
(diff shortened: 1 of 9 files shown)


ovirt-infra:

MoserMichael:
this is a test comment. please ignore.

MoserMichael:
this is a test comment

MoserMichael:
newcmt

cmd/manager/main.go
```diff
@@ -75,7 +75,7 @@ func main() {
 	ctx := context.TODO()
 
 	// Become the leader before proceeding
-	err = leader.Become(ctx, "node-maintenance-operator-lock")
```
dankenigsberg, line 78: I don't know much about NMO and it's leader election, feel free to silence me ;-)

Still, this namechange smells dangerous: and old copy of NMO may think that its the leader because it holds `node-maintenance-operator-lock`, but a new copy may already be the leader since it holds `openshift-kni-lifecycle-node-maintenance-operator-lock`.

slintes: you're right, kindly ignore this PR

slintes:
Running 2 operators in 2 namespaces isn't a problem, they just do the same without influencing each other. Existing leader election per namespace is good enough.

/close

PR #98, opened 2020-07-07
OLM integration

slintes:
The goal of this PR is to deploy NMO with OLM for e2e tests. Implemented in these steps:

- update to operator-sdk 0.18.2
- use dependencies according to a fresh project created by operator-sdk
- removed all unneeded replace directives
- cleanup with `go mod tidy && go mod vendor`
- smaller code fixes because of k8s / client-go API changes

The steps above resulted into a lot of modified files in the vendor dir. But sooner or later we need to update our dependencies, and IMHO now is fine, because NMO will be tested for the next release anyway.

- switch back to drain pkg of k8s.io/kubectl: because of a dependency mess we copied that package into our repo, this is reverted now

- added a make target (`gen-crds`) for updating CRD manifests
- updated operator's Dockerfile and scripts to latest version (like in a fresh operator-sdk project)
- big cleanup for csv generation:
  - removed unneeded and outdated scripts and resources and Dockerfiles
  - moved scripts from build to hack
  - updated CSV generation
  - added bundle and index image generation

- updated cluster-sync to install OLM if needed and deploy NMO with it

- fixed e2e test: the operator is deployed by make cluster-sync, but the operator-sdk's test framework
always wants to install it itself as well?! So don't use it, but a plain client-go k8s client

Replaces #84 partially. Webhook can be added in a follow up PR

```release-note
Install NMO with OLM in CI
```


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

slintes:
CI status:

k8s-1.17 is green :partying_face: 
k8s-1.13 does not exist anymore after kubevirtci update
ocp-4.3 does not get ready
Fix for the 2 latter: https://github.com/kubevirt/project-infra/pull/544

If someone knows how to sign off that commit produced by `make pull-ci-changes` for getting DCO green, please let me know. Whatever I tried was leading me into rebase hell :frowning:

slintes:
/hold

Separated dependency updates and drain package into #99

slintes:
/hold cancel

closed #99, doesn't work

slintes:
KubeVirtCI's ocp-4.4 also does not work (cluster not getting ready), so no way around using Openshift CI for OCP.
However, using OLM on the k8s lane is very valueable already and very close to what will happen on OCP :slightly_smiling_face:

Makefile
```diff
@@ -63,45 +83,52 @@ test:
...
 
-container-build: container-build-operator container-build-registry
+container-build: container-build-operator container-build-bundle container-build-index
 
 container-build-operator: csv-generator
 	docker build -f build/Dockerfile -t $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) .
 
-container-build-registry:
-	docker build -f build/Dockerfile.registry -t $(IMAGE_REGISTRY)/$(REGISTRY_IMAGE):$(IMAGE_TAG) .
+container-build-bundle:
```
beekhof, line 99: Not a prereq for this PR, but could we add a comment with links to what the bundle and index parts are for?
Helpful for those that come after us :)

build/Dockerfile.registry
```diff
@@ -1,15 +0,0 @@
...
-# drop all additional files.
-RUN find manifests/node-maintenance-operator/*/ -type f ! -name '*_crd.*' ! -name '*.clusterserviceversion.*' -exec rm -f {} \;
-
-# Initialize the database
-RUN initializer --manifests manifests/node-maintenance-operator/ --output bundles.db
-
-# There are multiple binaries in the origin-operator-registry
-# We want the registry-server
-ENTRYPOINT ["registry-server"]
-CMD ["--database", "bundles.db"]
```
beekhof, line 15: Does this relate to the old style OLM format thats being deprecated?
Would we have an reason to keep a cleaned up version of this around?

hack/sync.sh
```diff
@@ -16,6 +16,30 @@ fi
...
+# Deploy OLM if needed
+if [[ $KUBEVIRT_PROVIDER = k8s* ]]; then
+  SELF=$( realpath $0 )
+  BASEPATH=$( dirname $SELF )
+  . "${BASEPATH}/gen-operator-sdk.sh"
+
+  set +e
+  ${OPERATOR_SDK} olm status
+  if [[ $? != 0 ]] ; then
+    ${OPERATOR_SDK} olm install
```
beekhof: Wasn't there a whole script somewhere that was trying to do the same as this one line?

slintes: I don't think we installed OLM earlier, so no old script.

Makefile
slintes: fair point, the readme etc. probably needs an update anyway, forgot about it

build/Dockerfile.registry
slintes: This registry image is replaced by the new index image (plus bundle image), and they basically do the same (providing the CSV and CRD manifests), they are just build in different ways.
So I don't see a reason to keep this one.

slintes:
/override dco

ignoring the squashing commit by the kubevirtci update

PR #99, opened 2020-07-08
Update deps

slintes:
Separated from #98 

- update to operator-sdk 0.18.2
- use dependencies according to a fresh project created by operator-sdk
- removed all unneeded replace directives
- cleanup with go mod tidy && go mod vendor
- smaller code fixes because of k8s / client-go API changes
- switch back to drain pkg of k8s.io/kubectl: because of a dependency mess we copied that package into our repo, this is reverted now

```release-note
Update dependencies and revert to drain package of k8s.io/kubectl
```


slintes:
/close

doesn't work, more changes from #98 are needed to make this work, and then separation does not make sense anymore :man_shrugging:

PR #83, opened 2020-05-31
if failed to drain pods then requeue the reconcile request after timeout

MoserMichael:
returning an error will lead to bounded exponential wait between
reconcile requests. That is not desirable for repeated draining
attempts.

Signed-off-by: Michael Moser <mmoser@redhat.com>

PR #100, opened 2020-07-15
Fix requeue on drain error

slintes:
When draining the node fails, Requeue with a fixed delay instead of the default exponential delay.

Replaces #83

And some cleanup:
- use k8s ptr funcs for getting a pointer of something
- use Go's time / duration everywhere instead of vars named someTimeInSeconds
- removed unneeded helper funcs
- handle nil values and MicroTime in `ExpectEqualWithNil` and use it at more places.

```release-note
Fix requeue delay on drain error
```


test/e2e/nodemaintenance_test.go
```diff
@@ -328,15 +346,13 @@ func  enterAndExitMaintenanceMode(t *testing.T) error {
...
-func nodeMaintenanceTest(t *testing.T) error {
+func TestNodeMaintenance(t *testing.T) {
 
-	for i:=0; i < testIterations; i+=1 {
-		if err := enterAndExitMaintenanceMode(t); err != nil {
-			t.Fatalf("failed to enter maintenance mode. error %v", err);
-		}
+	if err := enterAndExitMaintenanceMode(t); err != nil {
+		showDeploymentStatus(t, nil)
+		t.Fatalf("failed to enter maintenance mode. error %v", err)
```
beekhof, line 353: nit: s/enter/enter or exit/

slintes: done

PR #101, opened 2020-07-20
Fixed make check

slintes:
In the previous version `make check` didn't fail when source files have wrong formatting, or generators didn't run.
Now all formatters and linters overwrite files, and `make check` calls all generators, and fails when it finds modified but uncommitted files. Also a new formatter for shell scripts was added.

The actual changes are in the 1st and 3rd commit, the newly formatted files in the 2nd and 4th.

```release-note
NONE
```


slintes:
/approve

@beekhof ready for review

PR #102, opened 2020-07-21
Added validation webhook

### LINKED ISSUE #58: Patching/Updating a NodeMaintenance CR leaves old node unschedulable
On creation of a NodeMaintenance CR, a specific node becomes unschedulable.
If an existing NodeMaintenance CR is updated and `nodeName` is changed to a different node name, the old node would stay unschedulable after deletion

How to reproduce:
1. Deploy NMO as it states in the main repository page
2. Apply the following NodeMaintenance CR:
```
apiVersion: kubevirt.io/v1alpha1
kind: NodeMaintenance
metadata:
  name: nodemaintenance-xyz
spec:
  nodeName: node02
  reason: "Test node maintenance"
```
3. Check that the node became unschedulable:
```
$ kubectl get nodes
NAME     STATUS                     ROLES    AGE   VERSION
node01   Ready                      master   14m   v1.16.2
node02   Ready,SchedulingDisabled   <none>   14m   v1.16.2
node03   Ready                      <none>   14m   v1.16.2
node04   Ready                      <none>   13m   v1.16.2
```
4. Patch the NodeMaintenance CR and change `nodeName`:
```
$ kubectl patch nodemaintenance nodemaintenance-xyz --type=json -p '[{"op":"replace", "path":"/spec/nodeName", "value":"node03"}]'
```
5. Validate that `node03` is still schedulable:
```
$ kubectl get nodes
NAME     STATUS                     ROLES    AGE   VERSION
node01   Ready                      master   27m   v1.16.2
node02   Ready,SchedulingDisabled   <none>   27m   v1.16.2
node03   Ready                      <none>   27m   v1.16.2
node04   Ready                      <none>   26m   v1.16.2
```
6. Delete the NodeMaintenance CR:
```
$ kubectl delete nodemaintenance nodemaintenance-xyz
nodemaintenance.kubevirt.io "nodemaintenance-xyz" deleted
```
7. Notice that `node02` is still unschedulable:
```
$ kubectl get nodes
NAME     STATUS                     ROLES    AGE   VERSION
node01   Ready                      master   29m   v1.16.2
node02   Ready,SchedulingDisabled   <none>   28m   v1.16.2
node03   Ready                      <none>   28m   v1.16.2
node04   Ready                      <none>   28m   v1.16.2
```

Proposed solution:
Prevent any patches or updates to a NodeMaintenance CR.

### LINKED ISSUE #97: Handle CR updates in vaildating webhook
In #84 an webhook is introduced for validating the creation of nodemaintenance CRs.
We should also validate updates to CRs.

- updates to the nodeName field: should probably be forbidden (are such updates handled at all today?)
- updates to the reason field: should probably be allowed. Check if something needs to be done for handling such updates

wdyt? @beekhof @MoserMichael

slintes:
For validation of CRs during creation and updates we need a validation webhook.

Replaces #84 
Fixes #97 
Fixes #58 

Validations implemented in this PR are:
- on creation:
  - validate nodename, a node with that name has to exist
  - prevent multiple CRs for the same node
- on update:
  - prevent nodename change

Interesting implementation details:
- Added `deploy/webhooks/nodemaintenancd.webhook.yaml`, will be added to CSVs by `make generate-bundle`
- Restrict `installMode` to `AllNamespaces` only, and modified `OperatorGroup` accordingly. Otherwise OLM will restrict webhook calls to CRs in namespaces of the OperatorGroup, which doesn't work for cluster scoped CRDs.
- Added code for serving the webhook
- Workaround for OLM installation: patch olm-operator to `latest` image because of unreleased bugfix
- Added ctual validation code
- Added unit and e2e test

```release-note
Added webhook for nodename validation
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ fmt: whitespace goimports
 
 .PHONY: goimports
 goimports:
-	go run golang.org/x/tools/cmd/goimports -w ./pkg ./cmd
+	go run golang.org/x/tools/cmd/goimports -w ./pkg ./cmd ./test
 
 .PHONY: whitespace
 whitespace: $(all_sources)
@@ -48,7 +48,7 @@ whitespace: $(all_sources)
 
 .PHONY: vet
 vet: $(cmd_sources) $(pkg_sources)
-	go vet -mod=vendor ./pkg/... ./cmd/...
+	go vet -mod=vendor ./pkg/... ./cmd/... ./test/...
 
 .PHONY: verify-unchanged
 verify-unchanged:
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -167,6 +167,8 @@ func setupWebhookServer(mgr manager.Manager) error {
 
 	server.Register("/validate-nodemaintenance-kubevirt-io-v1beta1-nodemaintenances", admission.ValidatingWebhookFor(&v1beta1.NodeMaintenance{}))
 
+	v1beta1.InitValidator(mgr.GetClient())
+
 	return nil
 
 }
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
 
 	"github.com/spf13/pflag"
@@ -16,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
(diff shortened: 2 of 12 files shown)


slintes:
WIP
/hold

slintes:
Oh, the validation webhook is called now, but the CR not processed anymore... :thinking:
Will investigate tomorrow.

pkg/webhooks/serve.go
```diff
@@ -0,0 +1,48 @@
...
+	WebhookCertDir  = "/apiserver.local.config/certificates"
+	WebhookCertName = "apiserver.crt"
+	WebhookKeyName  = "apiserver.key"
+)
+
+var log = logger.Log.WithName("webhook")
+
+func Serve(mgr manager.Manager, stop <-chan struct{}) error {
+
+	// Make sure the certificates are mounted, this should be handled by the OLM
```
beekhof, line 25: We were looking in the wrong place before?

slintes: no, when I mentioned yesterday that certs are not mounted yet I missed that the webhook's deploymentName [0] has to match the operator's deploment name in the CSV, and the former is derived from the service name in the webhook template [1]. So all good, it was just work in progress.

[0] https://github.com/kubevirt/node-maintenance-operator/pull/102/files#diff-0e416e752895627bc1a91e19c6be6e3cR200
[1] https://github.com/kubevirt/node-maintenance-operator/pull/102/files#diff-0132d87c0cf6c4433c9fcac6a4a769bdR12

slintes:
/hold cancel
/approve

@beekhof ready for review

PR #107, opened 2020-07-29
Added master quorum validation

slintes:
For the master quorum validation we check the PodDisruptionBudget of the etcd-quorum-guard. When it allows disruptions, we can
put a master node into maintenance.

This works on Openshift only, on k8s clusters (not having the etcd-quorum-guard) this validation will be skipped.

For e2e tests on k8s a fake PDB will be created.

Replaces #76 

```release-note
Added master quorum validation
```

### DIFF
--- a/deploy/webhooks/nodemaintenance.webhook.yaml
+++ b/deploy/webhooks/nodemaintenance.webhook.yaml
@@ -28,4 +28,4 @@ webhooks:
       - v1beta1
       # - v1 enable this as soon as we vendor a version of controller runtime which understands it
     sideEffects: None
-    timeoutSeconds: 5
+    timeoutSeconds: 15
--- a/test/e2e/nodemaintenance_test.go
+++ b/test/e2e/nodemaintenance_test.go
@@ -145,24 +145,25 @@ func checkInvalidLease(t *testing.T, nodeName string) error {
 	return nil
 }
 
-func countNodes(t *testing.T) (int, string, error) {
+func getNodes(t *testing.T) ([]string, []string, error) {
+	masters := make([]string, 0)
+	workers := make([]string, 0)
+
 	nodesList := &corev1.NodeList{}
 	err := Client.List(context.TODO(), nodesList, &client.ListOptions{})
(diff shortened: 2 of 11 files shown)


slintes:
CI is green :)

/approve

deploy/webhooks/nodemaintenance.webhook.yaml
```diff
@@ -28,4 +28,4 @@ webhooks:
       - v1beta1
       # - v1 enable this as soon as we vendor a version of controller runtime which understands it
     sideEffects: None
-    timeoutSeconds: 5
+    timeoutSeconds: 15
```
beekhof, line 31: You hit some timeouts?

test/e2e/nodemaintenance_test.go
```diff
@@ -558,7 +618,7 @@ func checkFailureStatus(t *testing.T) {
 	}
 }
 
-func waitForDeployment(t *testing.T, namespace, name string, replicas int, retryInterval, timeout time.Duration) error {
+func waitForDeployment(t *testing.T, namespace, name string, retryInterval, timeout time.Duration) error {
```
beekhof, line 621: We never want deployments with more than a single copy of the Pod?

deploy/webhooks/nodemaintenance.webhook.yaml
slintes: yes, but tbh they might have been caused by misconfiguration of RBAC in an earlier stage. However, the default 30 seconds, so with 15 we are still really fast ;)

test/e2e/nodemaintenance_test.go
slintes: It wasn't used, and I I didn't come up with a usecase... so I went the KISS way :shrug:

slintes:
> enterAndExitMaintenanceMode() is (and was already) very long... maybe its time to split it up into multiple tests?

oh yes, it's on my todo list, switching to ginkgo/gomega and give it some more structure :+1:

PR #115, opened 2020-08-11
Fix e2e test

slintes:
Running the e2e test on OpenshiftCI revealed some issues:
- putting master nodes into maintenance at (nearly) the same time won't be prevented by the master quorum validation, because the etcd-quorum-guard PDB isn't updated fats enough.
- the lease was expected in the wrong namespace
- seems the webhook setup is slower, so wait a bit after NMO deployment (update to a better solution when https://github.com/kubernetes-sigs/controller-runtime/issues/723 is fixed)

### DIFF
--- a/hack/sync-deploy.sh
+++ b/hack/sync-deploy.sh
@@ -108,6 +108,10 @@ until [[ $success -eq 1 ]] || [[ $iterations -eq $max_iterations ]]; do
     else
         # All resources deployed successfully
         success=1
+        # Fixme Webhook setup is slow... service needs to be created, endpoint needs to be created, and webhook server must be running
+        # Didn't find an easy but good solution yet, so just wait a bit for now
+        echo "[INFO] Giving the webhook some time to setup"
+        sleep 30s
     fi
     set -e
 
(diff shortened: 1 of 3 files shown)


slintes:
/hold cancel

let's see if CI is green, if so: ready to be merged

hack/sync-deploy.sh
```diff
@@ -108,6 +108,10 @@ until [[ $success -eq 1 ]] || [[ $iterations -eq $max_iterations ]]; do
     else
         # All resources deployed successfully
         success=1
+        # Fixme Webhook setup is slow... service needs to be created, endpoint needs to be created, and webhook server must be running
+        # Didn't find an easy but good solution yet, so just wait a bit for now
+        echo "[INFO] Giving the webhook some time to setup"
+        sleep 30s
```
beekhof, line 114: A timeout is fine for now, but perhaps creating/deleting an invalid CRD in a loop until it fails would tell you when it's finally set up?

slintes: good idea :+1:

PR #117, opened 2020-08-25
Switch from leader-for-life to leader-with-lease

slintes:
The latter works better with evicted pods, and since NMO actively evicts
pods, there is some chance that it evicts itself as well.
See https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#leader-election

```release-note
Switch from leader-for-life to leader-with-lease
```


beekhof:
Looks like the build failed but I'm definitely onboard with the concept

slintes:
> Looks like the build failed but I'm definitely onboard with the concept

just code formatting, fixed, thx for review :+1:

PR #121, opened 2020-09-01
Remove webhook workaround, not needed anymore

slintes:
Webhook works now for cluster scoped CRD when being installed in ownNamespace mode.

```release-note
NONE
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -10,12 +10,9 @@ import (
 
 	"github.com/spf13/pflag"
 
-	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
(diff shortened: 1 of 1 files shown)


beekhof:
Can we get some additional comments in the commit regarding what changed to allow us to remove the work-around?
Presumably this was something in OLM?  Mentioning which version might be good.

slintes:
> Can we get some additional comments in the commit regarding what changed to allow us to remove the work-around?
> Presumably this was something in OLM? Mentioning which version might be good.

@beekhof Fair point... so I tried to find out what changed:

- no change in OLM, it still sets the namespace selector
- k8s changed behaviour [0], it ignores namespace selectors for cluster scoped resources... but: that change is years old. So it explains why it works, but not why it didn't work before :thinking: Tbh I applied the workaround from HCO without checking if it is (still) needed, but I'm sure that I was running into some kind of problem with webhooks earlier. Maybe I misunderstood the problem at that time, and it was fixed by other changes. I don't have a better explanation, and maybe HCO is fixing another problem with it.

I just verified that CI uses the ownNamespace installMode, and that the namespaceSelector is set on the webhook config by OLM, and the validation tests are green. So I'm confident we really don't need this workaround... and it probably never was needed at all.

[0] https://github.com/kubernetes/kubernetes/commit/0d717272ccbec12189946325594a8a5f29d6bd68

slintes:
> and maybe HCO is fixing another problem with it

Reading through a bz related to another issue I saw that
- HCO was indeed trying to solve another issue with the workaround (being able to validate namespaced CRs in other namespaces
- and that OLM is reconciling the webhook config these days, so that the WA causes other troubles now

So I likely misunderstood the webhook issue when I was running into it.

beekhof:
thanks for the extra info
/lgtm

slintes:
KubeVirtCI is drunk, it complains about unused imports which were removed bu this PR.
Will rebase...

slintes:
> KubeVirtCI is drunk, it complains about unused imports which were removed by this PR.
> Will rebase...

git was drunk, merge issue, should be fixed now
thx KubevirtCI for retriggering tests when other PRs were merged :+1:

PR #131, opened 2020-10-06
Reject maintenance requests for unhealthy nodes

n1r1:
Reject maintenance create requests for node associated with unhealthy machine.
unhealthy machine = machine which meets the unhealthy criteria in machine healthcheck controller (in MAO), and only if the MHC CR is annotated with `host.metal3.io/external-remediation`.

```release-note
Reject maintenance requests for node associated with unhealthy machine.
```


pkg/apis/nodemaintenance/v1beta1/nodemaintenance_validator_test.go
```diff
@@ -3,6 +3,8 @@ package v1beta1
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	machinev1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
```
slintes, line 6: nit picking: we usually only have core Go packages in 1st block of imports, can you move this is the next to next block please?

pkg/apis/nodemaintenance/v1beta1/nodemaintenance_validator_test.go
```diff
@@ -118,6 +124,92 @@ var _ = Describe("NodeMaintenance Validation", func() {
...
+				})
+
+				It("should not be rejected", func() {
+					nm := getTestNMO(existingNodeName)
+					err := nm.ValidateCreate()
+					Expect(err).ToNot(HaveOccurred())
+				})
+			})
+
+			Context("node with invalid machine ref format should be rejected", func() {
```
slintes, line 186: "should be rejected" is in the "It()" description and can be removed here

pkg/apis/nodemaintenance/v1beta1/nodemaintenance_validator_test.go
```diff
@@ -118,6 +124,92 @@ var _ = Describe("NodeMaintenance Validation", func() {
...
+				})
+				It("should be rejected", func() {
+					nm := getTestNMO(existingNodeName)
+					err := nm.ValidateCreate()
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring(InvalidMachineFormat, invalidMachineRef))
+				})
+			})
+
+			Context("node with invalid machine ref format", func() {
```
slintes, line 200: I think this should be "node with non existent machine"?

pkg/apis/nodemaintenance/v1beta1/nodemaintenance_validator_test.go
```diff
@@ -118,6 +124,92 @@ var _ = Describe("NodeMaintenance Validation", func() {
...
+				BeforeEach(func() {
+					node.Annotations = make(map[string]string)
+					node.Annotations[MachineRefAnnotation] = "foo/bar"
+				})
+
+				It("node with non existent machine should not be rejected", func() {
```
slintes, line 206: and here just "should not be rejected"

slintes (changes requested):
Nice work, just some nits, found just some nits :+1: 
The OpenShift e2e test failed though, maybe the RBAC role must be extended for `machines`?

n1r1:
after a quick discussion we realized that a better solution would be that the remediation controller will try to obtain the lease, just as NMO does.

/close

PR #135, opened 2020-11-01
Use leader election to handle node leases

n1r1:


n1r1:
closing until we get a decision on the right path to move forward with
/close

PR #136, opened 2020-11-15
[WIP] PoC - Use lease lock from client-go 

n1r1:
Use lease lock from client-go instead of directly interact with the lease object, also take a different approach for how the controller interacts with the lease package which is safer in terms of lease semantics

i.e. make sure we reconcile before the lease is expired to renew it etc.


n1r1:
closing until we get a decision on the right path to move forward with 
/close

PR #139, opened 2020-12-01
Dockerize "make check" command

slintes:
Needed for moving the check job from Travis to KubeVirtCI


slintes:
and this will make the new jobs mandatory automatically: https://github.com/kubevirt/project-infra/pull/753 (needs approval from a KubevirtCI engineer)

PR #140, opened 2020-12-01
Bug 1895033: Remove master node affinity

slintes:
CNV uses the subscription to set a node selector, which conflicts with the node affinity.

```release-note
Removed master node affinity
```


slintes:
this is based on #139, so merge that one first

/hold

slintes:
#139 is in
/hold cancel

PR #150, opened 2021-01-07
Fixed .dockerignore

slintes:
It broke operator version reporting because of missing .git dir, so git decribe didn't work

```release-note
NONE
```


slintes:
something is terrible broken.... looking tomorrow...

slintes:
This will open a rabbit hole...
- verify-unchanged currently doesn't work because of missing .git
- running `make check` with docker build doesn't modify local files, which is a bad dev experience

Will open a new PR with a complete fix without backporting to release branch.

/close

PR #151, opened 2021-01-08
Fix dockerized code check

slintes:
- Use explicit ignore for .idea, otherwise .git will be ignored as well which breaks several things like version info and modified file check
- With the former `check.Dockerfile` and `docker build` for running code checks, potential file changes weren't written back to the local file system, which makes it unusable for local development. So we use docker run with volume mapping now
- Updated package / CSV versions
- introduced tools/tools.go for dev dependencies, and `go run ...` to run them from the vendor dir instead of downloading them on every CI run

```release-note
Fixed and improved dockerized code check
```


slintes:
/approve

@mshitrit can you have a look please and lgtm?

PR #153, opened 2021-02-02
Update the kubectl dep to pull in a fix which causes high CPU usage

n1r1:
Update the kubectl dep to pull in a fix which causes high CPU usage when draining a node with Pods containing a PodDisruptionBudget. 

upstream fix: https://github.com/kubernetes/kubernetes/pull/95260

```release-notes
Update the kubectl dep to pull in a fix which causes high CPU usage when draining a node with Pods containing a PodDisruptionBudget. 

upstream fix: https://github.com/kubernetes/kubernetes/pull/95260
```


n1r1:
cc @slintes

PR #159, opened 2021-05-25
Prepare next minor release

slintes:
After the just released version v0.9.0, set the stage for v0.10.0


slintes:
/hold cancel
needs to be merged before Ci because of the update 4.9 csv channel

PR #163, opened 2021-08-31
Prepare v0.11.0 for CNV / OCP 4.10

slintes:
Signed-off-by: Marc Sluiter <msluiter@redhat.com>


slintes:
openshift-e2e is expected to fail because of new channel

slintes:
> openshift-e2e is expected to fail because of new channel

see https://github.com/openshift/release/pull/21482

PR #164, opened 2021-09-24
Operator SDK update

slintes:
Short title, long story: in order to be future proof and getting rid of old technical debt the project is re-created using the latest operator-sdk (versiom 1.12.0).

The process in simple steps was:
- delete all but e2e tests
- create project
- merge code and config back into new project

For the detailed steps see commit history.
Last commit adds vendor dir, reviewers might want to checkout the code and skip that commit when reviewing... :) 

Things being left out while merging back:
- must gather image: re-add this only when needed
- e2e test on KubeVirtCI: this was using a very old version, and adds much complexity. Not sure anymore if it's worth it, or if e2e on OpenShiftCI is enough?
- old manifests: they were needed in an older OLM package format, today not anymore. For getting older versions just check git history
- deployment manifests: use `operator-sdk run bundle ...` instead

TODOs in a follow up:
- remove e2e job config from KubeVirtCI and then the `automation` dir (it's a no-op now)
- update Dockerfile locations in OpenshiftCI and then remove `build` dir
- test git tag based upstream release
- make sure HCO still works with this (it should, see relevant commit)
- update downstream build config

```release-note
Re-created project with latest operator-sdk 1.12.0 and get rid of technical debt
```


test/e2e/clients.go
```diff
@@ -8,8 +8,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
-	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
-	nmoapi "kubevirt.io/node-maintenance-operator/pkg/apis"
+	//operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	nmoapi "kubevirt.io/node-maintenance-operator/api/v1beta1"
```
rgolangh, line 12: just to make sure - the move to beta is intentional?

test/e2e/clients.go
```diff
@@ -29,10 +29,10 @@ func init() {
 		klog.Exit(err.Error())
 	}
 
-	// in case we need to work with OLM types
-	if err := operatorsv1alpha1.AddToScheme(scheme.Scheme); err != nil {
-		klog.Exit(err.Error())
-	}
+	//// in case we need to work with OLM types
```
rgolangh, line 32: I didn't see this removed later. maybe remove if not needed.

test/e2e/clients.go
rgolangh: maybe this is the new generated paty

slintes: yes, it's just a changed package, API version was v1beta1 before as well

test/e2e/clients.go
slintes: done, thx (also the one above this)

controllers/nodemaintenance_controller.go
```diff
@@ -55,16 +72,379 @@ type NodeMaintenanceReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
 func (r *NodeMaintenanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
```
rgolangh, line 86: in master its 
`reqLogger := log.WithFields(log.Fields{"Request.Namespace": request.Namespace, "Request.Name": request.Name})`

and then keeps using that to the rest of the scope.

controllers/nodemaintenance_controller.go
```diff
@@ -55,16 +72,379 @@ type NodeMaintenanceReconciler struct {
...
+	r.logger = logger
```
rgolangh, line 87: I think using both a logger per scope and mutating the inner logger is misleading.
maybe `r.logger = log.withFields...` and then only `r.logger` in all places?

controllers/nodemaintenance_controller.go
rgolangh: will log.FromContext(ctx) do the same?

controllers/nodemaintenance_controller.go
```diff
@@ -55,16 +72,379 @@ type NodeMaintenanceReconciler struct {
...
+
+func (r *NodeMaintenanceReconciler) setOwnerRefToNode(instance *nodemaintenancev1beta1.NodeMaintenance, node *corev1.Node) {
+
+	for _, ref := range instance.ObjectMeta.GetOwnerReferences() {
+		if ref.APIVersion == node.TypeMeta.APIVersion && ref.Kind == node.TypeMeta.Kind && ref.Name == node.ObjectMeta.GetName() && ref.UID == node.ObjectMeta.GetUID() {
+			return
+		}
+	}
+
+	// logger is nil... r.logger.Info("setting owner ref to node")
```
rgolangh, line 302: need to resole this

rgolangh: but of course not holding this pr. maybe just leave a todo. your call

controllers/nodemaintenance_controller.go
```diff
@@ -55,16 +72,379 @@ type NodeMaintenanceReconciler struct {
...
+	r.logger.Info("Reconciling with exponential duration")
+	return reconcile.Result{}, err
+}
+
+func (r *NodeMaintenanceReconciler) onReconcileError(nm *nodemaintenancev1beta1.NodeMaintenance, err error) (reconcile.Result, error) {
+	return r.onReconcileErrorWithRequeue(nm, err, nil)
+
+}
+
+//// set readable timestamp format in logs, not reference time
```
rgolangh, line 444: cleanup?

controllers/nodemaintenance_controller.go
```diff
@@ -55,16 +72,379 @@ type NodeMaintenanceReconciler struct {
...
+//func init() {
+//	log.SetFormatter(&log.TextFormatter{TimestampFormat: "2006-01-02 15:04:05.000000", FullTimestamp: true})
+//}
+
+// writer implements io.Writer interface as a pass-through for klog.
```
rgolangh, line 449: kudos for pushing less interesting code down.

controllers/nodemaintenance_controller.go
```diff
@@ -55,16 +72,379 @@ type NodeMaintenanceReconciler struct {
...
+type writer struct {
+	logFunc func(args ...interface{})
+}
+
+// Write passes string(p) into writer's logFunc and always returns len(p)
```
rgolangh, line 454: is this being used? doesn't look like related to the controller package.

controllers/suite_test.go
```diff
@@ -71,10 +80,35 @@ var _ = BeforeSuite(func() {
...
 	Expect(k8sClient).NotTo(BeNil())
 
-}, 60)
+	k8sManager, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:             scheme.Scheme,
+		MetricsBindAddress: "0",
+	})
+	Expect(err).ToNot(HaveOccurred())
+
+	// comment in when moving to "normal" testEnv test
```
rgolangh, line 90: comment needed?

controllers/suite_test.go
```diff
@@ -52,13 +55,19 @@ func TestAPIs(t *testing.T) {
 var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
+	// call start or refactor when moving to "normal" testEnv test
```
rgolangh, line 59: why is start/stopTestEnv needed rather that Before/AfterSuite?

automation/test.sh
```diff
@@ -1,32 +1,4 @@
...
-
-# Make sure that the VM is properly shut down on exit
-trap '{ make cluster-down; }' EXIT SIGINT SIGTERM
-
-make cluster-down
-make cluster-up
-make cluster-sync
-make cluster-functest
+echo "KubeVirtCI based e2e tests not supported anymore, check OpenshiftCI jobs"
+echo "TODO Update CI config and remove this"
```
rgolangh, line 4: :+1:

rgolangh: maybe a link to the openshift/release repo with the ci configs?

bundle/manifests/node-maintenance-operator.clusterserviceversion.yaml
```diff
@@ -240,5 +240,6 @@ spec:
       - nodemaintenances
     sideEffects: None
     targetPort: 9443
+    timeoutSeconds: 20
```
rgolangh, line 323: I think this was 15 on master?

controllers/nodemaintenance_controller.go
slintes: good question, and yes it does (didn't check where the magic happens though)
`2021-10-01T12:10:30.962962014Z 2021-10-01T12:10:30.962Z	INFO	controller-runtime.manager.controller.nodemaintenance	Reconciling NodeMaintenance	{"reconciler group": "nodemaintenance.kubevirt.io", "reconciler kind": "NodeMaintenance", "name": "nodemaintenance-test-maintenance", "namespace": ""}`

controllers/nodemaintenance_controller.go
slintes: fair point, but I go with `r.logger = log.FromContext(ctx)`

controllers/nodemaintenance_controller.go
slintes: forgot about this one, thx for the heads up. Seems the logger was only missing during unit tests, so I just added it

controllers/nodemaintenance_controller.go
slintes: yes, it is used: it implements the writer interface for the struct above by passing the bytes to a logger. The struct is initialized and used for the out and error streams of the "drainer", see bottom of `initDrainer` func
Does it belong here? Probably not. As other code also doesn't. But I didn't want to do more refacorings than needed for the sdk update in this PR 🤷🏼‍♂️

controllers/suite_test.go
slintes: well... at least for me it has a meaning 😉 Since it doesn't hurt, I leave it

controllers/suite_test.go
slintes: I start and stop for every test, see BeforeEach / AfterEach in `controller_test.go`. Lazy way of cleaning test resources after every test and keeping the test code nearer to the original version, where the old fake client was re-initialized for every test

automation/test.sh
slintes: a link would be useless, I plan to delete this in a follow up, after removing the job from KubeVirtCI which calls this

bundle/manifests/node-maintenance-operator.clusterserviceversion.yaml
slintes: good catch. I was experimenting with this value while trying to get tests green

PR #188, opened 2023-10-25
[release-0.11] Update golang.org/x/net to 0.17.0

mhenriks:
Address CVE-2023-39325 and CVE-2023-44487

Requires updating golang to 1.17


razo7:
>Address https://github.com/advisories/GHSA-4374-p667-p6c8 and https://github.com/advisories/GHSA-qppj-fm5r-hxr3
Requires updating golang to 1.17

AFAIK Golang v1.17 doesn't include any fixes for the HTTP/2 rapid reset CVEs. 
I think to address these CVEs NMO should patch `golang.org/x/net to 0.17.0`, and` k8s.io dependencies for 1.25 - 1.28` and maybe even disable HTTP/s on the webhook server. The Golang version might be bumped even to v1.21.3 or v1.20.10 which should include a patch. But all of these recommendations might be a huge change for one PR :)

>/go/pkg/mod/go.uber.org/zap@v1.26.0/internal/pool/pool.go:40:6: too many errors
note: module requires Go 1.19

Looking at the logs it seems like you have updated some dependency that requires Golang v1.19. 
It might be easier to investigate the changes if you can separate the large commit into a few commits/changes and then a last commit for updating modules using `make go-tidy go-vendor`

mhenriks:
@razo7 thanks for the comments.  Since there are no planned releases upstream we don't need to update the compiler to 1.20.10.  Just whatever minimally works.  Patch compiler will of course be needed for downstream release in order to address the CVE

razo7:
Looks better.
/approve
/lgtm

There is no e2e test for this branch and only to [`master`](https://github.com/openshift/release/tree/master/ci-operator/config/kubevirt/node-maintenance-operator). So I am holding in case you want more tests.
/hold

mhenriks:
@razo7 thanks for the review!  I was not planning to add any functional tests since no functionality changed.

